PR #164, opened 2020-06-24
Download the prebuilt cctools from S3

JCash:


server/docker-base/Dockerfile
```diff
@@ -102,6 +97,11 @@ RUN \
   rm -rf apple-libtapi && \
   rm -rf cctools-port
 
+# from version 1.2.171
+RUN wget -q -O - ${DM_PACKAGES_URL}/cctools-port-darwin14-3f979bbcd7ee29d79fb93f829edf3d1d16441147-linux.tar.gz | tar xz -C /usr/local --strip-components=1
+
+ENV LD_LIBRARY_PATH /usr/local/tapi1.4/lib:${LD_LIBRARY_PATH}
+
```
JCash, line 104: Since the vanilla engine now can use prebuilt cctools, there's no need for us to build it here as well.

PR #170, opened 2020-08-21
Add a parallel compilation for C/C++ files

aglitchman:
My game has a lot of C++ source files. Extender compiles C++ files one at a time, so every small change in the source code requires about 40-50 seconds to build. That's a lot of time!

This pull request adds a parallel compilation for C/C++ files, and the code uses all available CPU cores/threads. I suppose that it should utilise a single thread by default if only you didn't specify ENV variable with a different setting.

These changes are an example of how this can be done. If you decide to accept this pull request, I will polish it according to your requirements.

**Now Extender builds the game in 10 seconds on my 8-core PC.

### DIFF
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -17,6 +17,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


JCash (approved):
I think it looks good!
This gives me an idea on how I want this to move forward with threaded support.
We'll keep it as-is, and improve upon it if/when we need to.
(i..e make it more reusable for the other steps)

Thanks!

PR #177, opened 2020-09-26
Use aapt2 to compile android resources and generate R.java files

britzl:
This will allow us to send back the compiled resources to the client for use in the AAB build while at the same time getting R.java files with resource ids exactly matching the generated resources.

### DIFF
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -10,6 +10,8 @@ import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.error.YAMLException;
 
+import java.io.BufferedWriter;
+import java.io.FileWriter;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/defold/extender/TreePrinter.java
```diff
@@ -0,0 +1,59 @@
+package com.defold.extender;
```
britzl, line 1: This was useful while testing to quickly get an overview of the folder structure when processing or building android packages. We can remove it and I'll put it in a gist for future use.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -740,6 +742,122 @@ public boolean accept(File file) {
...
+    private static File createDir(String parent, String child) throws IOException {
+        return createDir(new File(parent), child);
+    }
+
+
+    private Map<String, Object> createContext() throws ExtenderException {
+        HashMap<String, Object> empty = new HashMap<>();
+        Map<String, Object> context = context(empty);
+        return context;
+    }
```
britzl, line 761: We did this in a lot of places. Might as well move it to a small helper function

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -740,6 +742,122 @@ public boolean accept(File file) {
...
+            String command = templateExecutor.execute(platformConfig.aapt2linkCmd, context);
+            processExecutor.execute(command);
+        }
+        catch (IOException | InterruptedException e) {
+            throw new ExtenderException(e, processExecutor.getOutput());
+        }
+
+        return files;
+    }
+
```
JCash, line 860: We should move the android functions (later) to an android specific module, since the Extender.java is really starting to be quite large.

server/src/main/java/com/defold/extender/TreePrinter.java
JCash: I think it's ok to have here.
Fyi, I generally use the `tree` command on the command line.
(I.e. not deleting the job folder, then logging into the server)

JCash (approved):

PR #182, opened 2020-11-20
Use the Android Gradle plugin to migrate dependencies to Android X

britzl:
Migration to Android X will happen if the `jetifier` property is set in the app.manifest:

```
    armv7-android:
        context:
            excludeLibs: []
            excludeJars: []
            excludeSymbols: []
            symbols: []
            libs: []
            linkFlags: []
            jetifier: true
```

### DIFF
--- a/server/docker-base/template.build.gradle
+++ b/server/docker-base/template.build.gradle
@@ -20,7 +20,7 @@ allprojects {
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 29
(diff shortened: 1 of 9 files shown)


server/docker-base/template.build.gradle
```diff
@@ -1,27 +1,41 @@
-apply plugin: 'java'
+buildscript {
```
britzl, line 1: Install the Android Gradle plugin

JCash (approved):
Looks good

PR #186, opened 2021-02-24
dx -> d8 migration

filosov:
Replace dx invocation by d8. Remove -source 1.6 -target 1.6 from javac command. Upgrade android build tools to 30.0.3.


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1332,25 +1332,14 @@ private File buildMainDexList(List<String> jars) throws ExtenderException {
...
         context.put("mainDexList", mainDexList.getAbsolutePath());
 
         String command = platformConfig.dxCmd;
 
-        // Until it's part of the build.yml
-        if (command.indexOf("--main-dex-list") == -1) {
-            command = command.replace("--multi-dex" , "--multi-dex --main-dex-list={{mainDexList}}");
-        }
-
         command = templateExecutor.execute(command, context);
```
JCash, line 1343: This removal is ok since we haven't needed since 1.2.163 and that version has been discontinued, given our 6 month backwards compatibility policy (for native extensions)

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1332,25 +1332,14 @@ private File buildMainDexList(List<String> jars) throws ExtenderException {
...
-
-        // The empty list is also present for backwards compatability with older build.yml
-        List<String> empty_list = new ArrayList<>();
 
         Map<String, Object> context = createContext();
-        context.put("classes_dex", classesDex.getAbsolutePath());
         context.put("classes_dex_dir", buildDirectory.getAbsolutePath());
         context.put("jars", jars);
-        context.put("engineJars", empty_list);
         context.put("mainDexList", mainDexList.getAbsolutePath());
```
JCash, line 1339: I still believe we need these engine jars in order to include the source in the app.

JCash (changes requested):
I'm not convinced that the "engineJars" should be removed.

server/src/main/java/com/defold/extender/Extender.java
filosov: If I understood everything correctly "engineJars" always passed as empty list to templateExecutor. All engine jars are collected by method **getAllJars** and would be passed as part of "jars" variable.

JCash: I admit it does look confusing.
However, the command line outputs this (I posted the current command line in the other PR):
>dx -JXms2048m -JXmx2048m --dex --multi-dex --main-dex-list=/tmp/job6102724887021115144/build/main_dex_list.txt --output /tmp/job6102724887021115144/build /dynamo_home//ext/share/java/android-support-multidex.jar /dynamo_home//share/java/glfw_android.jar /dynamo_home//share/java/gamesys_android.jar /dynamo_home//share/java/sound_android.jar /tmp/job6102724887021115144/build/tmp0rjava/R.jar

You can see the command line outputs from the server by running it with 
> DM_DEBUG_COMMANDS=1 ./server/scripts/run-local.sh

JCash: Ah, I see now.
It's because `getAllJars()` was modified in a previous release, to return the union of the two lists, thus making the second occurrance of `engineJars` redundant.

JCash (approved):

PR #196, opened 2021-04-08
Update users at runtime

britzl:


JCash (approved):

README_SECURITY.md
```diff
@@ -68,3 +68,5 @@ This defines two users: "bob" and "may". Bob has permission to create Windows, L
 Authentication is performed using standard Basic access authentication. The authentication data can be sent as an `Authorization` request header, but that is inconvenient when using the command line tools (bob.jar) or the Defold editor. The username and password can be sent as part of the build server URL set in the Preferences window of the editor and using the `--build-server` option to bob.jar:
 
     java -jar bob.jar --build-server https://bob:super5ecret@myextender.com
+
+It is also possible to specify a username and password in the environment variables DM_EXTENDER_USERNAME and DM_EXTENDER_PASSWORD.
```
britzl, line 72: This was added in a previous PR for those uncomfortable with adding the username and password in the URL itself. We forgot to add documentation for this though.

.gitmodules
```diff
@@ -1,3 +1,7 @@
 [submodule "server/scripts/shared"]
 	path = server/scripts/shared
 	url = git@github.com:defold/aws-build-tools.git
+
+[submodule "server/scripts/task-definitions"]
```
britzl, line 5: The AWS task definitions are here. These now contain application.yml overrides with paths to resources we do not wish to share openly.

server/src/main/java/com/defold/extender/WebSecurityConfig.java
```diff
@@ -21,16 +19,7 @@
...
-	Resource usersResource;
-
-	private Properties loadUsers() throws Exception {
-		Properties users = new Properties();
-		if (usersResource != null) {
-			users.load(usersResource.getInputStream());
-		}
-		return users;
-	}
+	private InMemoryUserDetailsManager userDetailsManager = new InMemoryUserDetailsManager();
```
britzl, line 22: We start with an empty user details manager. It will get updated before the first build via the UserUpdateService.

server/src/main/java/com/defold/extender/WebSecurityConfig.java
```diff
@@ -69,7 +58,11 @@ protected void configure(HttpSecurity http) throws Exception {
 
 	@Override
 	protected void configure(AuthenticationManagerBuilder auth) throws Exception {
-		Properties users = loadUsers();
-		auth.userDetailsService(new InMemoryUserDetailsManager(users));
+		auth.userDetailsService(userDetailsManager);
+	}
+
+	@Bean
```
britzl, line 64: Expose the user details manager as a Spring Bean so that we can access it from the UserUpdateService

server/src/main/java/com/defold/extender/services/UserUpdateService.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Value("${extender.authentication.users}")
+    Resource usersResource;
+
+    @Value("${extender.authentication.update-interval}")
+    long updateInterval;
+
+    private long lastUpdateTimestamp = 0;
+
+    private Properties loadUsers() {
```
britzl, line 32: Load the users from the user resource (file or url)

README_SECURITY.md
```diff
@@ -38,8 +38,27 @@ docker run ... -e extender.authentication.platforms=windows,linux,macos extender
 ```
 
 ### Users
-Users are listed in a text file, one username per line followed by the user password, user roles (platform access) and user status ("enabled" or "disabled"). The build script will copy the files in `extender/server/users` to the Docker image when building. You can specify which user file to use in `application.yml` or by passing the `extender.authentication.users` environment variable when starting Docker.
+Users are listed in Java properties format with one username per line followed by the user password, user roles (platform access) and user status ("enabled" or "disabled"). Example:
 
+```
+bob = super5ecret,ROLE_WINDOWS,ROLE_LINUX,ROLE_MACOS,enabled
+may = top5ecret,ROLE_MACOS,enabled
```
JCash, line 45: The "superSecret" and "topSecret" is a bit unclear I think.
Perhaps "password1" and "password2" is better?

README_SECURITY.md
```diff
@@ -54,17 +73,26 @@ The same by passing an environment variable when launching Docker:
...
-Example user file:
+#### Using a URL
+Example using `application.yml` where users are loaded from the content server from `https://www.mysite.com/extender-users`:
 
 ```
-bob = super5ecret,ROLE_WINDOWS,ROLE_LINUX,ROLE_MACOS,enabled
-may = top5ecret,ROLE_MACOS,enabled
+extender:
+    authentication:
+        users: https://www.mysite.com/extender-users
```
JCash, line 82: Perhaps make it clearer that it expects a text file?
In our task definition we use an explicit url to a users.txt.

README_SECURITY.md
```diff
@@ -54,17 +73,26 @@ The same by passing an environment variable when launching Docker:
...
+```
+
+The same by passing an environment variable when launching Docker:
+
+```
+docker run ... -e extender.authentication.users=https://www.mysite.com/extender-users extender/extender;
```
JCash, line 88: I don't see a change for run-local.sh?
How does this work with the current workflow? (I.e. when I build for Switch)

JCash (changes requested):

README_SECURITY.md
britzl: Well it doesn't actually have to be a file. It does an HTTP GET so anything that serves a response in the correct format is fine. It could be a REST API that grabs the users from a DB.

This can perhaps be clarified though.

README_SECURITY.md
britzl: I'll move the extender.authentication.platforms config to the task definition.

server/src/main/java/com/defold/extender/services/UserUpdateService.java
```diff
@@ -0,0 +1,81 @@
...
+        try {
+            users.load(usersResource.getInputStream());
+        }
+        catch(IOException e) {
+            LOGGER.info("UserUpdateService - Unable to update users. " + e.getMessage());
+        }
+        return users;
+    }
+
+    // user = password,ROLE_ANDROID,ROLE_IOS,ROLE_HTML5,ROLE_WINDOWS,ROLE_LINUX,ROLE_MACOS,ROLE_SWITCH,disabled
```
JCash, line 47: Is this a "standard" format?
If we later need to add/remove info to this (not sure what, but let's say email), will it be easy to do? Will it be easy to remove fields?
Not sure we need to, just making sure the format is as solid we can make it/want it.

JCash (approved):

server/src/main/java/com/defold/extender/services/UserUpdateService.java
britzl: Yeah, so the InMemoryUserDetailsManager accepts a Properties instance as one of its constructors:

https://docs.spring.io/spring-security/site/docs/4.2.20.RELEASE/apidocs/org/springframework/security/provisioning/InMemoryUserDetailsManager.html#InMemoryUserDetailsManager-java.util.Properties-

I can't seem to find the example of how it is used anymore.

BUT since we aren't creating the InMemoryUserDetailsManager with a Properties instance in the constructor we're actually free to change the format if we like.

server/src/main/java/com/defold/extender/services/UserUpdateService.java
```diff
@@ -0,0 +1,81 @@
...
+    private void updateUsers(Properties users) {
+        for(String username : users.stringPropertyNames()) {
+            List<String> userSettings = Arrays.asList(users.getProperty(username).split(","));
+            final String password = userSettings.get(0);
+            final boolean disabled = userSettings.get(userSettings.size() - 1).equals("disabled");
+            final String[] authorities = userSettings.subList(0, userSettings.size() - 1).toArray(new String[0]);
```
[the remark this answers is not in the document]
britzl: What do you mean by this? Yes, we update/create users with a set of authorities/roles and a password.

server/src/main/java/com/defold/extender/services/UserUpdateService.java
```diff
@@ -0,0 +1,99 @@
...
+    Resource usersResource;
+
+    @Value("${extender.authentication.update-interval}")
+    long updateInterval;
+
+    private long lastUpdateTimestamp = 0;
+
+    /**
+     * Load users from the resource specified by extender.authentication.users
+     * inb application.yml or env variable.
```
JCash, line 38: inb

server/src/main/java/com/defold/extender/services/UserUpdateService.java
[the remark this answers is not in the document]
JCash: My question is, is "authorities[0] == password" ?
Or am I reading it wrong?

server/src/main/java/com/defold/extender/services/UserUpdateService.java
JCash: 👍 
It's good that we can change it.

Workflow question: Will it be easy?
I'm thinking of the update workflow here.
Let's say we update the parsing function, since we want to add/remove a field.
We then need to build stage server, and point it to a separate users.txt.
During these two weeks (the beta period), we need to update two separate users.txt to keep them working for our users.
Then, after two weeks, we can merge the stage/users.txt to the prod/users.txt.

I'm not going to push for any changes at this point though.since this is yet theoretical and not an actual use case yet.

server/src/main/java/com/defold/extender/services/UserUpdateService.java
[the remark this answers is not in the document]
britzl: Oh, good catch! Yes indeed, the password was included as one of the authorities.

server/src/main/java/com/defold/extender/services/UserUpdateService.java
britzl: Yeah, I get what you are saying. What we do know is that password comes first, then a list of authorities/roles all prefixed with ROLE_ and finally a value indicating if the user is enabled or not. We're not locked into a format which we cannot change at this moment, but I agree that it makes sense to pick a more extensible format.

JCash (approved):

PR #197, opened 2021-04-16
Issue 5725: Added support for commonsrc+pluginsrc folders

JCash:
Can build .proto files into engine + plugin source
Can generate a plugin .jar file

### DIFF
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -19,7 +19,6 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.*;
-import java.util.concurrent.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -36,6 +35,7 @@ class Extender {
     private final File uploadDirectory;
     private final File jobDirectory;
     private final File buildDirectory;
+    private final PlatformConfig commonPlatformConfig;
     private final PlatformConfig platformConfig;
     private final ExtensionManifestValidator manifestValidator;
     private final TemplateExecutor templateExecutor = new TemplateExecutor();
@@ -53,12 +53,18 @@ class Extender {
     private List<File> gradlePackages;
     private int nameCounter = 0;
 
+
+    static final String FOLDER_ENGINE_SRC = "src";      // source for the engine library
+    static final String FOLDER_PLUGIN_SRC = "pluginsrc";// source for the pipeline/format plugins
+    static final String FOLDER_COMMON_SRC = "commonsrc";// common source shared between both types
+
     static final String APPMANIFEST_BASE_VARIANT_KEYWORD = "baseVariant";
     static final String APPMANIFEST_WITH_SYMBOLS_KEYWORD = "withSymbols";
     static final String APPMANIFEST_JETIFIER_KEYWORD = "jetifier";
     static final String FRAMEWORK_RE = "(.+)\\.framework";
     static final String JAR_RE = "(.+\\.jar)";
     static final String JS_RE = "(.+\\.js)";
+    static final String PROTO_RE = "(?i).*(\\.proto)";
     static final String ENGINE_JAR_RE = "(?:.*)\\/share\\/java\\/[\\w\\-\\.]*\\.jar$";
 
     private static final String MANIFEST_IOS    = "Info.plist";
@@ -145,7 +151,10 @@ class Extender {
         this.platform = platform;
         this.sdk = sdk;
 
+        // TODO: Add a way to merge these platform configs: common -> platform -> arch-platform
+        this.commonPlatformConfig = config.platforms.get("common");
         this.platformConfig = getPlatformConfig(platform);
+
         this.appManifestContext = ExtenderUtil.getAppManifestContext(appManifest, platform, baseVariantManifest);
         LOGGER.info("Using context for platform: " + platform);
 
@@ -161,7 +170,17 @@ class Extender {
                 envContext.put("env." + sysEnvEntry.getKey(), sysEnvEntry.getValue());
             }
 
-            Set<String> keys = this.platformConfig.env.keySet();
+            Set<String> keys;
+            if (this.commonPlatformConfig != null) {
+                keys = this.commonPlatformConfig.env.keySet();
+                for (String k : keys) {
+                    String v = this.commonPlatformConfig.env.get(k);
+                    v = templateExecutor.execute(v, envContext);
+                    processExecutor.putEnv(k, v);
+                }
+            }
+
+            keys = this.platformConfig.env.keySet();
             for (String k : keys) {
                 String v = this.platformConfig.env.get(k);
                 v = templateExecutor.execute(v, envContext);
@@ -320,6 +339,9 @@ class Extender {
     private Map<String, Object> context(Map<String, Object> manifestContext) throws ExtenderException {
         Map<String, Object> context = new HashMap<>(config.context);
 
+        if (commonPlatformConfig != null) {
+            context = Extender.mergeContexts(context, commonPlatformConfig.context);
+        }
         context = Extender.mergeContexts(context, platformConfig.context);
         context = Extender.mergeContexts(context, manifestContext);
 
@@ -429,7 +451,7 @@ class Extender {
         return allLibJars;
     }
 
-    private File compileFile(int index, File extDir, File src, Map<String, Object> manifestContext, List<String> commands) throws IOException, InterruptedException, ExtenderException {
+    private List<String> getIncludeDirs(File extDir) {
         List<String> includes = new ArrayList<>();
         File extIncludeDir = new File(extDir, "include");
         includes.add( ExtenderUtil.getRelativePath(jobDirectory, extIncludeDir ) );
@@ -438,10 +460,15 @@ class Extender {
         // Add the other extensions include folders
         for (File otherExtDir : this.getExtensionFolders()) {
             File otherIncludeDir = new File(otherExtDir, "include");
-            if (extIncludeDir.equals(otherIncludeDir))
+            if (extIncludeDir.equals(otherIncludeDir)) // don't add the current extension's include directory
                 continue;
             includes.add( ExtenderUtil.getRelativePath(jobDirectory, otherIncludeDir ) );
         }
+        return includes;
+    }
+
+    private File addCompileFileCpp_Internal(int index, File extDir, File src, Map<String, Object> manifestContext, String cmd, List<String> commands) throws IOException, InterruptedException, ExtenderException {
+        List<String> includes = getIncludeDirs(extDir);
 
         File o = new File(buildDirectory, String.format("%s_%d.o", src.getName(), index));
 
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -438,10 +460,15 @@ private File compileFile(int index, File extDir, File src, Map<String, Object> m
...
             File otherIncludeDir = new File(otherExtDir, "include");
-            if (extIncludeDir.equals(otherIncludeDir))
+            if (extIncludeDir.equals(otherIncludeDir)) // don't add the current extension's include directory
                 continue;
             includes.add( ExtenderUtil.getRelativePath(jobDirectory, otherIncludeDir ) );
         }
+        return includes;
+    }
+
+    private File addCompileFileCpp_Internal(int index, File extDir, File src, Map<String, Object> manifestContext, String cmd, List<String> commands) throws IOException, InterruptedException, ExtenderException {
```
JCash, line 470: The old name "compileFile" was now a misnomer, since it only sets up the command.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -478,66 +513,144 @@ private File compileMain(File maincpp, Map<String, Object> manifestContext) thro
...
+                List<File> _srcFiles =  new ArrayList<>(FileUtils.listFiles(srcDir, null, true));
+                _srcFiles = Extender.filterFiles(_srcFiles, regEx);
+
+                // sorting makes it easier to diff different builds
+                Collections.sort(_srcFiles, NameFileComparator.NAME_INSENSITIVE_COMPARATOR);
+                srcFiles.addAll(_srcFiles);
+            }
+        }
+        return srcFiles;
+    }
```
JCash, line 529: Some restructuring.
Could probably me moved to ExtenderUtil.java

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -478,66 +513,144 @@ private File compileMain(File maincpp, Map<String, Object> manifestContext) thro
...
+    private List<File> generateProtoCxxForEngine(File extDir, Map<String, Object> manifestContext, List<File> protoFiles) throws IOException, InterruptedException, ExtenderException {
+        List<File> generated = new ArrayList<>();
 
-            // sorting makes it easier to diff different builds
-            srcFiles = _srcFiles.toArray(new File[_srcFiles.size()]);
+        Map<String, Object> tmpcontext = context(manifestContext);
 
+        if (commonPlatformConfig == null || commonPlatformConfig.protoEngineCxxCmd == null) {
+            LOGGER.info("Not supporting .proto files");
+            return generated;
```
JCash, line 543: support old versions of our dmsdk, which doesn't support this.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -478,66 +513,144 @@ private File compileMain(File maincpp, Map<String, Object> manifestContext) thro
...
+
+        String suffix = "";
+        if (language.equals("cpp"))
+            suffix = ".pb.cc";
+        else if (language.equals("java"))
+            suffix = ".java";
+        else if (language.equals("python"))
+            suffix = "_pb2.py";
+        else
+            throw new ExtenderException("Proto output doesn't support language: " + (language!=null?language:"null"));
```
JCash, line 591: These files are for manipulating source/output content.
We can convert .proto into files for cpp, java and python.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -478,66 +513,144 @@ private File compileMain(File maincpp, Map<String, Object> manifestContext) thro
...
+        return generated;
+    }
+
+
+
+    private void buildExtension(File manifest, Map<String, Object> manifestContext) throws IOException, InterruptedException, ExtenderException {
+        File extDir = manifest.getParentFile();
+
+        // Gather all the C++ files
+        File[] srcDirs = { new File(extDir, FOLDER_COMMON_SRC), new File(extDir, FOLDER_ENGINE_SRC) };
```
JCash, line 626: Look in the `commonsrc/` + `src/` folders

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -478,66 +513,144 @@ private File compileMain(File maincpp, Map<String, Object> manifestContext) thro
...
+        File[] srcDirs = {srcDir};
+        return listFiles(srcDirs, regEx);
+    }
 
-        File[] srcFiles = null;
-        if (srcDir.isDirectory()) {
-            Collection<File> _srcFiles = new ArrayList<>();
-            _srcFiles = FileUtils.listFiles(srcDir, null, true);
-            _srcFiles = Extender.filterFiles(_srcFiles, platformConfig.sourceRe);
+    private List<File> generateProtoCxxForEngine(File extDir, Map<String, Object> manifestContext, List<File> protoFiles) throws IOException, InterruptedException, ExtenderException {
```
JCash, line 536: This function converts .proto files into _ddf.h/_ddf.cpp, that are used by the engine to read the files at runtime.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -478,66 +513,144 @@ private File compileMain(File maincpp, Map<String, Object> manifestContext) thro
...
-                throw (IOException)e.getCause();
-            } else if (e.getCause() instanceof InterruptedException) {
-                throw (InterruptedException)e.getCause();
-            } else {
-                throw new ExtenderException(e.getCause().toString());
-            }
-        } finally {
-            executor.shutdown();
-        }
+        ProcessExecutor.executeCommands(processExecutor, commands); // in parallel
```
JCash, line 653: Moved the parallelism into the ProcessExecutor itself

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -553,6 +666,119 @@ public Void call() throws Exception {
...
+        // ***************************************************************************
+        // C++
+        {
+            List<File> srcFiles = listFiles(srcDirs, platformConfig.sourceRe);
+
+            if (srcFiles.isEmpty()) {
+                LOGGER.info("No C++ source found for plugin. Skipping {}", extDir);
+            } else {
+                // We leave the C++ protobuf support for a later date
+                List<File> generatedFiles = new ArrayList<>();//generateProtoSrcForPlugin(extDir, manifestContext, protoFiles, "cpp");
```
JCash, line 694: Currently disabled until later.
We want to be able to use C++ code to transform data (e.g. textures)

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -553,6 +666,119 @@ public Void call() throws Exception {
...
+                // Collect all classes into a Jar file
+                Map<String, Object> context = createContext();
+                File outputJar = new File(extBuildDir, String.format("plugin%s.jar", manifestContext.get("extension_name")));
+
+                context.put("outputJar", outputJar.getAbsolutePath());
+                context.put("classesDir", classesDir.getAbsolutePath());
+                String command = templateExecutor.execute(commonPlatformConfig.jarCmd, context);
+                processExecutor.execute(command);
+
+                outputFiles.add(outputJar);
```
JCash, line 764: List the output files that are to be sent back to the client.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1426,19 +1652,18 @@ private void loadManifests() throws IOException, ExtenderException {
...
-            List<String> symbols = new ArrayList<>();
-
-            Set<String> keyset = manifestConfigs.keySet();
-            String[] keys = keyset.toArray(new String[keyset.size()]);
-            Arrays.sort(keys);
-
-            for (String extensionSymbol : keys) {
-                symbols.add(extensionSymbol);
-
+            List<String> symbols = getSortedKeys(manifestConfigs.keySet());
```
JCash, line 1665: We sort them so that it's easier to compare builds when debugging

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1715,6 +1971,7 @@ else if (platform.contains("web")) {
             outputFiles.addAll(buildAndroid(platform));
         }
         outputFiles.addAll(buildEngine());
+        outputFiles.addAll(buildPipelinePlugin());
```
JCash, line 1974: Send back all generated plugins (currently .jar file and "_pb2.py" files)

server/src/main/java/com/defold/extender/ProcessExecutor.java
```diff
@@ -86,4 +87,34 @@ public void setCwd(File cwd) {
     public void putLog(String msg) {
         output.append(msg);
     }
+
+    public static void executeCommands(ProcessExecutor processExecutor, List<String> commands) throws IOException, InterruptedException, ExtenderException {
```
JCash, line 91: Moved from Extender.java

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -53,12 +53,18 @@
     private List<File> gradlePackages;
     private int nameCounter = 0;
 
+
+    static final String FOLDER_ENGINE_SRC = "src";      // source for the engine library
+    static final String FOLDER_PLUGIN_SRC = "pluginsrc";// source for the pipeline/format plugins
+    static final String FOLDER_COMMON_SRC = "commonsrc";// common source shared between both types
```
JCash, line 59: These names are now part of the api.
Are they good enough?

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -553,6 +666,119 @@ public Void call() throws Exception {
...
+                List<String> commands = new ArrayList<>();
+
+                // Compile C++ source into object files
+                int i = getAndIncreaseNameCount();
+                for (File src : srcFiles) {
+                    File o = addCompileFileCppShared(i, extDir, src, manifestContext, commands);
+                    objs.add(ExtenderUtil.getRelativePath(jobDirectory, o));
+                    i++;
+                }
+                ProcessExecutor.executeCommands(processExecutor, commands); // in parallel
```
Jhonnyg, line 708: Should the name of the function indicate parallelism somehow? So the comments aren't necessary

Jhonnyg (approved):
Looks good!

server/src/main/java/com/defold/extender/Extender.java
britzl: Is plugin the word we are going to use? When talking to users or when users discuss Defold I see the following three used to describe what we call a Library project in the documentation:

* (Native) extension
* library
* plugin
* dependency

And soon an extension project will also contain source code for the editor right? That folder would probably be named `editorsrc`?

What about `buildsrc`, `buildersrc` or `pipelinesrc`?

`commonsrc` is good I think. It could also be named `sharedsrc` but I believe we use `common` elsewhere to indicate shared things.

JCash: "plugin" was a collecting term for all plugins (bob _and_ editor)

"buildsrc" is too ambigous I think. All source is built.

"pipelinesrc" is better.

I don't think we'll want a separate "editorsrc". I consider that part still a "plugin"

JCash: I think it's good to separate the meanings of the the terms (as we already do).
I wouldn't want to call everything "extensions" or "libraries".

Also, I don't really consider them confusing.
But I agree that we should be more specific when referring to them

* (Defold) library
* Dependency is the link to a library
* (Native) extension - statically linked source code. Avaialble through the Defold Library system.
* Plugin (a separate file/bundle to add code)

JCash: I chose "pluginsrc" over "pipelinesrc" because it was shorter, but more importantly because "plugin" is a well known concept. But as you mention, it might be confusing for others.

PR #202, opened 2021-07-10
Issue 5920: Updated to iOS 14.5, macOS 11.3, Xcode 12.5

JCash:


server/build.gradle
```diff
@@ -13,7 +13,7 @@ plugins {
     id 'java'
     id 'idea'
     id 'eclipse'
-    id 'org.springframework.boot' version '1.5.17.RELEASE'
+    id 'org.springframework.boot' version '1.5.22.RELEASE'
```
JCash, line 16: I had some boot issues, and this seemed to help

server/docker-base/Dockerfile
```diff
@@ -314,20 +194,17 @@ ENV \
...
   WINDOWS_VCINSTALLDIR="${PLATFORMSDK_WIN32}/MicrosoftVisualStudio2019/VC/Tools/MSVC/${WINDOWS_MSVC_2019_VERSION}/" \
   WINDOWS_VSINSTALLDIR="${PLATFORMSDK_WIN32}/MicrosoftVisualStudio2019/" \
-  WindowsLibPath="${PLATFORMSDK_WIN32}/WindowsKits/8.1/References/CommonConfiguration/Neutral" \
-  WindowsSdkDir="${PLATFORMSDK_WIN32}/WindowsKits/8.1/" \
-  WindowsSDKLibVersion="winv6.3/" \
-  WindowsSDKVersion="/"
+  WindowsLibPath="${PLATFORMSDK_WIN32}/WindowsKits/10/References/${WINDOWS_SDK_10_VERSION}" \
+  WindowsSdkDir="${PLATFORMSDK_WIN32}/WindowsKits/10/" \
+  WindowsSDKLibVersion="${WINDOWS_SDK_10_VERSION}" \
+  WindowsSDKVersion="${WINDOWS_SDK_10_VERSION}"
```
JCash, line 207: Cleaned up the old Windows sdk's (and iOS sdk's)
Saved ~2GB from the container.

server/docker-base/Dockerfile
```diff
@@ -581,6 +435,82 @@ COPY template.build.gradle ${EXTENSION_BUILD_GRADLE_TEMPLATE}
 COPY template.gradle.properties ${EXTENSION_GRADLE_PROPERTIES_TEMPLATE}
 COPY template.local.properties ${EXTENSION_LOCAL_PROPERTIES_TEMPLATE}
 
+#
```
JCash, line 438: I moved it last since rebuilding gets a lot faster when I constantly change variables.

server/docker-base/Dockerfile
```diff
@@ -581,6 +435,82 @@ COPY template.build.gradle ${EXTENSION_BUILD_GRADLE_TEMPLATE}
...
+  ln -s ${PLATFORMSDK_DIR}/XcodeDefault${XCODE_12_VERSION}.xctoolchain/usr/lib/arc /usr/local/lib/arc
+
+RUN mkdir -p /usr/local/lib/clang/10.0.1/lib/darwin && \
+    ln -s /opt/platformsdk/XcodeDefault${XCODE_12_VERSION}.xctoolchain/usr/lib/clang/${XCODE_12_CLANG_VERSION}/lib/darwin/libclang_rt.asan_osx_dynamic.dylib \
+    /usr/local/lib/clang/10.0.1/lib/darwin/libclang_rt.asan_osx_dynamic.dylib
+
+# For backwards compatbility, we must also keep a copy of the library folder
+# Remove when 1.2.185 is the base line
+RUN ln -s /opt/platformsdk/XcodeDefault${XCODE_12_VERSION}.xctoolchain/usr/lib/clang/${XCODE_12_CLANG_VERSION} \
+    /opt/platformsdk/XcodeDefault${XCODE_12_VERSION}.xctoolchain/usr/lib/clang/12.0.0
```
JCash, line 510: We use the new XCode but older build.yml's refer to the 12.0.0 clang version, so we need to make a symlink.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -170,6 +170,10 @@
             for (Map.Entry<String, String> sysEnvEntry : System.getenv().entrySet()) {
                 envContext.put("env." + sysEnvEntry.getKey(), sysEnvEntry.getValue());
             }
+            // Make custom env variables available for the template execution below.
+            for (Map.Entry<String, String> envEntry : env.entrySet()) {
+                envContext.put("env." + envEntry.getKey(), envEntry.getValue());
+            }
```
JCash, line 176: Basically only for the tests.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -369,25 +377,53 @@ static boolean isListOfStrings(List<Object> list) {
             context.put(k, v);
         }
 
-        // Added to build.yml in 1.2.174 (setting for using Clang 9)
+        // Added to build.yml in 1.2.185 (setting for using isystem headers)
```
JCash, line 380: The upgrade workaround.
Since the iOS 14.5 has a different layout than the 14.0 version, and those build.yml's didn't have these settings.

server/docker-base/Dockerfile
```diff
@@ -581,6 +435,82 @@ COPY template.build.gradle ${EXTENSION_BUILD_GRADLE_TEMPLATE}
...
+
+# For package instructions, see <defold>/scripts/mobile/package_xcode_and_sdks.sh
+# and <defold>/scripts/mobile/package_cctools.sh
+
+ENV LD_LIBRARY_PATH ${LD_LIBRARY_PATH}
+
+# Install flow from  >=1.2.174
+ENV XCODE_12_VERSION      12.5
+ENV MACOS_10_15_VERSION   10.15
+ENV IOS_14_VERSION        14.5
```
JCash, line 450: By updating this version number, we make it so that all our versions that use `IOS_14_VERSION` will now use 14.5. I.e. the ones using 1.2.174 an newer.

I don't think we've discussed this approach before, but we should give it a though, since it might be both good and bad.
Good: That we can update the version for everyone.
Bad: We change things for users that might not suspect it

server/src/test/java/com/defold/extender/ExtenderTest.java
```diff
@@ -39,7 +39,23 @@ public void testExtender() throws IOException, InterruptedException, ExtenderExc
...
+        env.put("XCODE_12_VERSION", "12.5");
+        env.put("IOS_14_VERSION", "14.5");
+        env.put("LIB_TAPI_1_6_PATH", "/usr/local/tapi1.6/lib");
+        env.put("MACOS_11_VERSION", "11.3");
+        env.put("XCODE_12_CLANG_VERSION", "12.0.5");
+        env.put("SWIFT_5_VERSION", "5.0");
+        env.put("SYSROOT", "/opt/platformsdk/MacOSX11.3.sdk");
+        env.put("LD_LIBRARY_PATH", "/usr/local/tapi1.6/lib");
+
+        Extender extender = new Extender("x86_64-osx", sdk, jobDir, uploadDir, buildDir, env);
```
JCash, line 58: Missing setup in order to have a local (unit test) build.yml that more resembles the actual build.yml. I.e. the build.yml will be easier to update.

server/docker-base/Dockerfile
britzl: I wonder if there's any radical changes between the 14.x versions?

britzl (approved):
Lots of changes. Looks good as far as I can see!

PR #209, opened 2021-10-11
Reduce big memory allocations when sending and receiving data from a remote builder

britzl:


server/src/main/java/com/defold/extender/ExtenderController.java
```diff
@@ -186,10 +186,8 @@ public void buildEngine(HttpServletRequest request,
             if (remoteBuilderEnabled && isRemotePlatform(platform)) {
                 LOGGER.info("Building engine on remote builder");
 
-                final byte[] bytes = remoteEngineBuilder.build(uploadDirectory, platform, sdkVersion);
+                remoteEngineBuilder.build(uploadDirectory, platform, sdkVersion, response.getOutputStream());
```
britzl, line 189: The final output of a build is supposed to go to the `response.getOutputStream()` so we might as well pass that to the remote engine builder.

PR #213, opened 2021-11-19
Added possibility to override context properties in build.yml

JCash:


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -248,6 +249,15 @@ private PlatformConfig getPlatformConfig(String platform) throws ExtenderExcepti
         return platformConfig;
     }
 
+    private String getPlatformConfigProperty(String property) {
```
JCash, line 252: Allows us to get properties from the PlatformConfig class (read form build.yml)
E.g. "x86_64-win32" will take precedence over the "common" platform.

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -216,7 +209,7 @@ public boolean accept(File file) {
         if (manifest.platforms.containsKey("common")) {
             Object v = manifest.platforms.get("common").context.get(name);
             if( v != null ) {
-                if (!Extender.isListOfStrings((List<Object>) v)) {
+                if (!ExtenderUtil.isListOfStrings((List<Object>) v)) {
```
JCash, line 212: Weird circular dependency between this util file and the Extender.java. :/

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -287,10 +280,108 @@ static String getAppManifestContextString(AppManifestConfiguration manifest, Str
...
+
+            if (v1 != null && v2 != null && !v1.getClass().equals(v2.getClass())) {
+                throw new ExtenderException(String.format("Wrong manifest context variable type for %s: Expected %s, got %s: %s", k, v1.getClass().toString(), v2.getClass().toString(), v2.toString()));
+            }
+            if (v2 != null && v2 instanceof List && !ExtenderUtil.isListOfStrings((List<Object>) v2)) {
+                throw new ExtenderException(String.format("The context variables only support lists of strings. Got %s (type %s)", v2.toString(), v2.getClass().getCanonicalName()));
+            }
+
+            if (v1 != null && v2 != null && v1 instanceof List) {
+                //System.out.printf("merge op %s : %s  + %s  (merge: %s)\n", key, v1.toString(), v2.toString(), isMergeOp?"true":"false");
```
JCash, line 352: removed

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -287,10 +280,108 @@ static String getAppManifestContextString(AppManifestConfiguration manifest, Str
...
+
+                if (isMergeOp) {
+                    v1 = ExtenderUtil.mergeLists((List<String>) v1, (List<String>) v2);
+                } else {
+                    v1 = v2;
+                }
```
JCash, line 358: If the property name ends with "_replace", the else-path here is taken, and we keep the v2 value as-is.

If not, we use the backwards compatible way of merging the two lists v1 and v2.

server/src/main/java/com/defold/extender/PlatformConfig.java
```diff
@@ -11,7 +11,7 @@
     public String exePrefix; // deprecated
     public String exeExt; // deprecated
     public String writeLibPattern;
-    public String writeShLibPattern = new String();
+    public String writeShLibPattern;
```
JCash, line 14: Having it as "null" makes it easier to check for overrides between the platform config instances.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -239,6 +239,7 @@
     }
 
     private PlatformConfig getPlatformConfig(String platform) throws ExtenderException {
+        PlatformConfig commonPlatformConfig = config.platforms.get("common");
```
britzl, line 242: Not ussed?

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -248,6 +249,15 @@ private PlatformConfig getPlatformConfig(String platform) throws ExtenderExcepti
         return platformConfig;
     }
 
+    private String getPlatformConfigProperty(String property) {
+        String value = ExtenderUtil.get(platformConfig, property);
+        if (value == null)
+        {
+            value = ExtenderUtil.get(commonPlatformConfig, property);
```
britzl, line 256: `commonPlatformConfig` is missing in this scope

britzl: Ah, no it's a private member of the class. Sorry.

britzl (changes requested):
See note on unused variable

server/src/main/java/com/defold/extender/Extender.java
JCash: Yeah, I started out with one idea of "merging" these two classes into one (might do later on), but instead implemented the getPlatformConfigProperty() instead for now.

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -287,10 +280,106 @@ static String getAppManifestContextString(AppManifestConfiguration manifest, Str
...
+        return items;
+    }
+
+    private static final String MERGE_KEY_REPLACE = "_replace";
+    static private boolean isMergeOp(String name)
+    {
+        if (name.endsWith(MERGE_KEY_REPLACE))
+            return false;
+        return true;
+    }
```
JCash, line 304: Q: Do we have a better way than this? Better name?

britzl: Can't think of anything right now to be honest.

britzl (approved):

PR #214, opened 2021-12-01
Update Nintendo Switch sdk to 13.3.0

JCash:


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -942,12 +942,21 @@ private void getProjectPaths(Map<String, Object> mainContext, Map<String, Object
         context.put("extLibs", patchLibs((List<String>) context.get("extLibs")));
         context.put("engineLibs", patchLibs((List<String>) context.get("engineLibs")));
 
-        String command = templateExecutor.execute(platformConfig.linkCmd, context);
+        List<String> commands = platformConfig.linkCmds; // Used by e.g. the Switch platform
```
JCash, line 945: This is a change from the beta branch (already live on the beta server)

server/docker-base/Dockerfile
```diff
@@ -349,13 +349,16 @@ RUN \
 # Switch SDK
 #
 
-ENV NINTENDO_SDK_ROOT ${PLATFORMSDK_DIR}/nx
-ENV SWITCH_SDK_FILENAME nx64-sdk-10.4.1.tar.gz
+ENV NINTENDO_SDK_ROOT_13_3_0 ${PLATFORMSDK_DIR}/nx-13.3.0
```
JCash, line 352: It feels safer to keep version numbers in the variable, so we don't accidentally update for users that want to stay on a version for a while

server/docker-base/Dockerfile
```diff
@@ -349,13 +349,16 @@ RUN \
...
+# Backwards compatibility, since they're still using that env variable
+ENV NINTENDO_SDK_ROOT         ${NINTENDO_SDK_ROOT_13_3_0}
```
JCash, line 354: The exception to the previous comment. However, since 10.4.1 isn't supported anymore, we had to update anyways.

britzl (approved):
Looks good!

PR #217, opened 2021-12-14
Linux update

JCash:
Current status:

* Using Geadle 6.9, as our gradle docker build plugin doesn't support 7.x
* Extender jar file now became two files by default, had to do some tricks to get a fat jar
* Spring server won't start properly, complaining about "No auto configuration classes found in META-INF/spring.factories"

Untested container functionality:

* Updated to Java 17, how will Android builds go?
* Only installed winetricks, without vcrun2015 (i.e. vc_redist). How willwin32/switch builds go?
* Using clang 13, how will iOS builds go?


JCash:
Downgraded back to OpenJDK 11 due to issues with using Gradle 6.1.1 (in the container)
I tried updating to Gradle 6.9.1, but it seems something (groovy) is locked to version 6.1.
I'm wondering if it's something to do with the Android sdk?

britzl:
> I tried updating to Gradle 6.9.1, but it seems something (groovy) is locked to version 6.1.
> I'm wondering if it's something to do with the Android sdk?

Yes, probably. We use Android Gradle Plugin version 4.0.1 (see template.build_gradle) and it depends on a specific Gradle version:

https://developer.android.com/studio/releases/gradle-plugin?buildsystem=ndk-build#updating-gradle

(Also see note on JDK version on the same page)

JCash:
I'm assuming you meant [build_gradle.template](https://cs.android.com/android/platform/superproject/+/master:sdk/templates/build_gradle.template)?

britzl:
> I'm assuming you meant [build_gradle.template](https://cs.android.com/android/platform/superproject/+/master:sdk/templates/build_gradle.template)?

No, I meant this file: https://github.com/defold/extender/blob/dev/server/docker-base/template.build.gradle

JCash:
The branch now builds locally, and needs to be tested more thoroughly.
E.g. as the stand along server.

Other changes:
Removed support for Emscripten 1.3.16 (we use 2.0.11 since 1.2.178)

britzl (approved):

PR #220, opened 2022-01-02
Updated the spring multipart file size parameter to 2.6.1

JCash:


server/src/main/resources/application.yml
```diff
@@ -33,11 +33,11 @@ extender:
         update-interval: 900000
 
 spring:
-    http:
+    servlet:
```
JCash, line 36: I never saw this in the migration guides :/

I got upload issues when trying to build the rive extension, but not in my other smaller tests:(

server/src/main/resources/application.yml
```diff
@@ -33,11 +33,11 @@ extender:
...
         multipart:
-            enabled: false
-            max-file-size: 1024mb
-            max-request-size: 1024mb
+            enabled: true
+            max-file-size: 1024MB
+            max-request-size: 1024MB
```
JCash, line 40: Needs to be upper case now

server/src/main/resources/application.yml
```diff
@@ -33,11 +33,11 @@ extender:
         update-interval: 900000
 
 spring:
-    http:
+    servlet:
         multipart:
-            enabled: false
-            max-file-size: 1024mb
-            max-request-size: 1024mb
+            enabled: true
```
JCash, line 38: Not sure why it was disabled?

PR #225, opened 2022-04-01
Fix misprint in ios build endpoint name when configure web security.

ekharkunov-ms:
We can build ios for armv7 without authentication because there is misprint in endpoint name.


britzl (approved):
Nice catch!

JCash (approved):

PR #226, opened 2022-04-01
Remove dependency duplication. Now extender start inside Jetty not inside Tomcat.

ekharkunov-ms:
Because of dependency duplication extender starts inside Apache Tomcat, not inside Eclipse Jetty. These bug was introduced when spring boot were updated (https://github.com/defold/extender/commit/38d045ba2123eae178cc52eb6130cad5198c2637).
These fix remove dependency duplication. Now extender runs inside Jetty.


britzl (approved):
This is a really good find! We specifically try to catch a Jetty EOFException in the extender controller and I guess it won't happen if we're running in Tomcat and not Jetty :-)

JCash (approved):

PR #233, opened 2022-06-21
Prune libraries the same way as the main engine linking is doing

JCash:


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -520,6 +520,11 @@ private File linkCppShared(File extBuildDir, List<String> objs, Map<String, Obje
         Map<String, Object> env = new HashMap<>();
         getProjectPaths(context, env);
 
+        context.put("engineLibs", ExtenderUtil.pruneItems(ExtenderUtil.getStringList(context, "engineLibs"), ExtenderUtil.getStringList(context, "includeLibs"), ExtenderUtil.getStringList(context, "excludeLibs")) );
+        context.put("engineJsLibs", ExtenderUtil.pruneItems(ExtenderUtil.getStringList(context, "engineJsLibs"), ExtenderUtil.getStringList(context, "includeJsLibs"), ExtenderUtil.getStringList(context, "excludeJsLibs")) );
+        context.put("objectFiles", ExtenderUtil.pruneItems(ExtenderUtil.getStringList(context, "objectFiles"), ExtenderUtil.getStringList(context, "includeObjectFiles"), ExtenderUtil.getStringList(context, "excludeObjectFiles") ));
+        context.put("dynamicLibs", ExtenderUtil.pruneItems(ExtenderUtil.getStringList(context, "dynamicLibs"), ExtenderUtil.getStringList(context, "includeDynamicLibs"), ExtenderUtil.getStringList(context, "excludeDynamicLibs")) );
```
JCash, line 526: This is the same way that the engine linker does it.

My dream is to simplify all these things a lot, but I still haven't found the time to do it yet.

AGulev (approved):
LGTM
how it's specified on the bob.jar side when you build lib? just as an app.manifest in project?

JCash:
You specify "--build-artifacts=plugins" to bob (and also the variant):
https://github.com/defold/extension-spine/blob/main/utils/build_plugins.sh#L76

JCash:
It it possibble to specify an app manifest as well, by passing in a custom .ini file which specifies an app manifest.
I didn't do it in this case though.

PR #236, opened 2022-08-16
Add support for asynchronous extender builds

britzl:
This adds asynchronous build support via the `/build_async/PLATFORM/VERSION` endpoint. Async builds return a job id which can be used to query build status via the `/job_status?jobId=JOBID` endpoint while the build is in progress. When a build has completed the build result can be retrieved through the `/job_result&jobId=JOBID` endpoint.


server/build.gradle
```diff
@@ -75,8 +80,32 @@ dependencies {
...
+    public static final String gitVersion = "${sha}";
+    public static final String buildTime = "${build_time}";
+}
+"""
+    def binding = ["sha" : sha, "build_time" : build_time]
+
+    def engine = new groovy.text.SimpleTemplateEngine()
+    def template = engine.createTemplate(text).make(binding)
+
+    new File(projectDir.toString() + "/src/main/java/com/defold/extender", "Version.java").text = template.toString()
```
JCash, line 108: What was the benefit of this as opposed to the python script?

client/src/main/java/com/defold/extender/client/ExtenderClient.java
```diff
@@ -47,6 +53,13 @@ public class ExtenderClient {
...
         this.extenderBaseUrl = extenderBaseUrl;
         this.cache = new ExtenderClientCache(cacheDir);
+        this.httpCookies = new BasicCookieStore();
+        this.buildSleepTimeout = Long.parseLong(System.getProperty("com.defold.extender.client.build-sleep-timeout", "5000"));
+        this.buildResultWaitTimeout = Long.parseLong(System.getProperty("com.defold.extender.client.build-wait-timeout", "240000"));
+    }
+
+    private void print(String s, Object... args) {
+        System.out.println(String.format(s, args));
     }
```
JCash, line 63: Should it print unconditionally?

client/src/main/java/com/defold/extender/client/ExtenderClient.java
```diff
@@ -141,6 +155,89 @@ private static Set<String> getCachedFiles(String json) throws ExtenderClientExce
...
+
+            final int statusCode = response.getStatusLine().getStatusCode();
+            if (statusCode == HttpStatus.SC_OK) {
+                String jobId = EntityUtils.toString(response.getEntity());
+                print("Async build request was accepted as job %s", jobId);
+                long currentTime = System.currentTimeMillis();
+                Integer jobStatus = 0;
+                Thread.sleep(buildSleepTimeout);
+                while (System.currentTimeMillis() - currentTime < buildResultWaitTimeout) {
+                    print("Checking job status for job %s", jobId);
```
JCash, line 181: I consider this very verbose information.
I don't want us end up in a situation where we cannot see our warnings and errors due to the amount of other info.

JCash (approved):
A lot of code, but most it has been battle tested.

I think the logging bit can be improved though.

server/build.gradle
britzl: I was debating wether to change this or not. Removing the python script drops a dependency on python in favour of Java.

client/src/main/java/com/defold/extender/client/ExtenderClient.java
britzl: Good point. Changed to java.util.logging

client/src/main/java/com/defold/extender/client/ExtenderClient.java
britzl: Removed

PR #237, opened 2022-09-15
Added a safe constructor to the yaml parser

britzl:


server/src/main/java/com/defold/extender/ExtenderYamlSafeConstructor.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.yaml.snakeyaml.constructor.Construct;
+import org.yaml.snakeyaml.constructor.Constructor;
+import org.yaml.snakeyaml.nodes.Tag;
+
+class ExtenderYamlSafeConstructor extends Constructor {
+
+	public ExtenderYamlSafeConstructor() {
+		super();
+		Construct c = this.yamlConstructors.get(null); // ConstructYamlObject
```
britzl, line 11: I'm not entirely happy with this. The ConstructYamlObject is an inner protected class of the parent and I would prefer to create an instance of it here instead of relying on the fact that it was set in the parent class for the null key....

I'm getting "error: Constructor.ConstructYamlObject() has protected access in Constructor.ConstructYamlObject"

server/src/main/java/com/defold/extender/ExtenderYamlSafeConstructor.java
```diff
@@ -0,0 +1,23 @@
...
+
+		this.yamlConstructors.put(new Tag(Tag.PREFIX + "com.defold.extender.Configuration"), c);
```
britzl, line 13: Explicitly define the classes below as ok to create

server/src/main/java/com/defold/extender/ExtenderYamlSafeConstructor.java
```diff
@@ -0,0 +1,23 @@
...
+		this.yamlConstructors.put(new Tag(Tag.PREFIX + "com.defold.extender.ManifestConfiguration"), c);
+		this.yamlConstructors.put(new Tag(Tag.PREFIX + "com.defold.extender.AppManifestConfiguration"), c);
+		this.yamlConstructors.put(new Tag(Tag.PREFIX + "com.defold.extender.PlatformConfig"), c);
+		this.yamlConstructors.put(new Tag(Tag.PREFIX + "com.defold.extender.AppManifestPlatformConfig"), c);
+		this.yamlConstructors.put(new Tag(Tag.PREFIX + "com.defold.extender.ManifestPlatformConfig"), c);
+		this.yamlConstructors.put(new Tag(Tag.PREFIX + "com.defold.extender.WhitelistConfig"), c);
+
+		this.yamlConstructors.put(null, undefinedConstructor);
```
britzl, line 21: And any other class should not be allowed (see SafeConstructor)

JCash (approved):

PR #240, opened 2022-09-27
Darwin cleanup

JCash:


server/src/main/java/com/defold/extender/AsyncBuilder.java
```diff
@@ -113,7 +113,7 @@ public void asyncBuildEngine(MetricsWriter metricsWriter, String platform, Strin
 
     @Scheduled(fixedDelayString="${extender.job-result.cleanup-period:10000}")
     public void cleanUnusedResults() {
-        LOGGER.info("Clean result folder started.");
+        LOGGER.debug("Clean result folder started.");
```
JCash, line 116: This became too annoying when running the server locally.
And it doesn't really give us anything, we can assume it's already running.

britzl (approved):

PR #241, opened 2022-09-29
Added option to specify additional headers to extender client

britzl:
This PR should be considered in combination with https://github.com/defold/defold/issues/6411

### DIFF
--- a/client/build.gradle
+++ b/client/build.gradle
@@ -8,7 +8,7 @@ repositories {
 
 jar {
     baseName = 'extender-client'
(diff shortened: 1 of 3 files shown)


JCash (approved):
Do we want a test for this in the extender? E.g. like in AuthenticationTest.java

britzl:
> Do we want a test for this in the extender? E.g. like in AuthenticationTest.java

Yes, you're right, we want that!

JCash (approved):

PR #244, opened 2022-10-23
FIx potential deadlock in file creation

JCash:


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -267,12 +267,7 @@ private String getNameUUID() {
     }
 
     private File createBuildFile(String name) {
-        File file;
-        do {
-            file = new File(buildDirectory, name);
-        } while (file.exists());
-
-        return file;
+        return new File(buildDirectory, name);
```
JCash, line 270: Originally, the loop was there to make sure we didn't create a duplicate temp file.
Then, the loop also updated the name for each iteration.

Since we don't do it anymore, we would end up in an infinite loop if it somehow existed.

Also, our other functions already create a unique id when calling this function.

britzl: Oh, wow, that was bad!

PR #246, opened 2022-10-28
Send build server logs on error

### LINKED ISSUE #221: Send additional information to the client for server errors
Sometimes the build data can trigger an Internal Server Error. In such a scenario no information except the response code and error title is sent back to the client. It would help greatly with extension development if we could send some additional information from the extender log back to the client.

Successful builds result in a `.internal/cache/%platform%/build.zip`. This archive contains a log.txt with the full extender log.

Failed builds do not result in a build.zip (and thus no log).

Thought: We recently added support for asynchronous extender builds using our command line tools ([PR](https://github.com/defold/extender/pull/236)). Brief explanation: A build request is sent from the client to start a build. The response returns a job id. The job id is used to poll for build status and then later to retrieve the build result (ie the build.zip). We could return all log output from when polling and getting the final status. This output can be printed to the terminal (and then later to the editor console when async builds are supported also from the editor).

britzl:
Async builds will now return the full build log when requesting the `/job_result`. The request previously only returned the stack trace of the error. The result (stack trace and log) will be written to the log file on the client.

Fixes #221

### DIFF
--- a/server/src/main/java/com/defold/extender/AsyncBuilder.java
+++ b/server/src/main/java/com/defold/extender/AsyncBuilder.java
@@ -50,8 +50,8 @@ public class AsyncBuilder {
         if (extender == null) return;
         try {
             FileOutputStream fos = new FileOutputStream(file, true);
-            List<File> logs = extender.writeLogs();
-            for (File log : logs) {
+            File log = extender.writeLog();
+            if (log.exists()) {
                 FileInputStream fis = new FileInputStream(log);
                 fis.transferTo(fos);
                 fis.close();
--- a/server/src/main/java/com/defold/extender/AsyncBuilder.java
+++ b/server/src/main/java/com/defold/extender/AsyncBuilder.java
@@ -50,7 +50,7 @@ public class AsyncBuilder {
         if (extender == null) return;
         try {
             FileOutputStream fos = new FileOutputStream(file, true);
-            List<File> logs = extender.writeLogs()
+            List<File> logs = extender.writeLogs();
             for (File log : logs) {
                 FileInputStream fis = new FileInputStream(log);
                 fis.transferTo(fos);
--- a/server/src/main/java/com/defold/extender/AsyncBuilder.java
+++ b/server/src/main/java/com/defold/extender/AsyncBuilder.java
@@ -113,13 +113,13 @@ public class AsyncBuilder {
             Files.move(tmpResult.toPath(), targetResult.toPath(), StandardCopyOption.ATOMIC_MOVE);
         } catch(EofException e) {
             File errorFile = new File(resultDir, BuilderConstants.BUILD_ERROR_FILENAME);
-            writeExceptionToFile(e, errorFile);
             writeExtenderLogsToFile(extender, errorFile);
+            writeExceptionToFile(e, errorFile);
             LOGGER.error("Client closed connection prematurely, build aborted", e);
         } catch(Exception e) {
             File errorFile = new File(resultDir, BuilderConstants.BUILD_ERROR_FILENAME);
-            writeExceptionToFile(e, errorFile);
             writeExtenderLogsToFile(extender, errorFile);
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/defold/extender/AsyncBuilder.java
```diff
@@ -84,15 +113,13 @@ public void asyncBuildEngine(MetricsWriter metricsWriter, String platform, Strin
...
+            writeExceptionToFile(e, errorFile);
+            writeExtenderLogsToFile(extender, errorFile);
             LOGGER.error("Client closed connection prematurely, build aborted", e);
         } catch(Exception e) {
             File errorFile = new File(resultDir, BuilderConstants.BUILD_ERROR_FILENAME);
-            PrintWriter writer = new PrintWriter(errorFile);
-            e.printStackTrace(writer);
-            writer.close();
+            writeExceptionToFile(e, errorFile);
+            writeExtenderLogsToFile(extender, errorFile);
```
JCash, line 122: Thinking chronologically, should these be swapped around?

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -2048,7 +2048,7 @@ private List<File> buildManifests(String platform) throws ExtenderException {
         return out;
     }
 
-    private List<File> writeLog() {
+    List<File> writeLogs() {
```
JCash, line 2051: Why the rename, it has one log?

The "list" was only for the "addAll" below.
I suggest changing that to an "add(log)" instead.

server/src/main/java/com/defold/extender/AsyncBuilder.java
```diff
@@ -45,6 +45,34 @@ public AsyncBuilder(DefoldSdkService defoldSdkService,
         this.keepJobDirectory = System.getenv("DM_DEBUG_KEEP_JOB_FOLDER") != null && System.getenv("DM_DEBUG_JOB_FOLDER") == null;
         this.resultLifetime = jobResultLifetime;
     }
+
+    private void writeExtenderLogsToFile(Extender extender, File file) {
+        if (extender == null) return;
+        try {
+            FileOutputStream fos = new FileOutputStream(file, true);
+            List<File> logs = extender.writeLogs();
```
JCash, line 53: This is always a list of 1 file.

JCash (approved):

server/src/main/java/com/defold/extender/Extender.java
britzl: I was considering this as well. I'll change it now that you also point it out!

JCash (approved):

server/src/main/java/com/defold/extender/AsyncBuilder.java
```diff
@@ -63,25 +65,25 @@ private void writeExtenderLogsToFile(Extender extender, File file) {
...
     @Async
-    public void asyncBuildEngine(MetricsWriter metricsWriter, String platform, String sdkVersion, 
+    public void asyncBuildEngine(MetricsWriter metricsWriter, String platform, String sdkVersion,
             File jobDirectory, File uploadDirectory, File buildDirectory) throws IOException {
         String jobName = jobDirectory.getName();
         Thread.currentThread().setName(String.format("async-build-%s", jobName));
         File resultDir = new File(jobResultLocation.getAbsolutePath(), jobName);
         resultDir.mkdir();
-        Extender extender;
+        Extender extender = null;
```
JCash, line 86: Isn't it null by default in Java?

JCash (approved):

server/src/main/java/com/defold/extender/AsyncBuilder.java
britzl: Well, no, it is considered "uninitialized" which means that I get an error when trying to pass it as an argument to the helper function.

PR #248, opened 2022-11-06
Automatically create Android R.java files for all Gradle dependencies

britzl:
This change automatically adds each resolved Android Gradle dependency containing resources as an extra package when linking resources using `aapt2`. This reduces or completely removes the need to specify `aaptExtraPackages` in `ext.manifest`.

Fixes defold/defold#6047

### DIFF
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -1015,7 +1015,7 @@ class Extender {
                             .collect(Collectors.toList());
     }
 
-    private List<File> getAndroidResourceFolders(String platform) {
+    private List<String> getAndroidResourceFolders(String platform) {
         // New feature from 1.2.165
         File packageDir = new File(uploadDirectory, "packages");
         if (!packageDir.exists()) {
@@ -1051,14 +1051,8 @@ class Extender {
 
         return packageDirs.stream()
                         .filter(f -> f.isDirectory())
+                        .map(File::getAbsolutePath)
                         .collect(Collectors.toList());
-    }
-
-    private List<String> getAndroidResourceFoldersAsStrings(String platform) {
-        return getAndroidResourceFolders(platform)
-                                        .stream()
-                                        .map(File::getAbsolutePath)
-                                        .collect(Collectors.toList());
     }
 
 
@@ -1147,7 +1141,9 @@ class Extender {
                 extraPackages.addAll((List<String>)mergedAppContext.get("aaptExtraPackages"));
             }
             if (!extraPackages.isEmpty()) {
-                context.put("extraPackages", String.join(":", extraPackages));
+                String extraPackagesString = String.join(":", extraPackages);
+                context.put("extraPackages", extraPackagesString);
+                LOGGER.info("Extra packages {}", extraPackagesString);
             }
 
             File manifestFile = new File(buildDirectory, MANIFEST_ANDROID);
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1016,44 +1016,50 @@ private List<File> getAndroidJniFolders(String platform) {
     }
 
     private List<File> getAndroidResourceFolders(String platform) {
-            // New feature from 1.2.165
-            File packageDir = new File(uploadDirectory, "packages");
-            if (!packageDir.exists()) {
-                return new ArrayList<>();
+        // New feature from 1.2.165
```
britzl, line 1019: Indentation was off...

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1015,47 +1015,47 @@ private List<File> getAndroidJniFolders(String platform) {
...
+        packageDirs.addAll(gradlePackages.stream()
+                                         .map(f -> new File(f, "res"))
+                                         .collect(Collectors.toList()));
 
-            return packageDirs.stream()
-                            .filter(f -> f.isDirectory())
-                            .collect(Collectors.toList());
+        return packageDirs.stream()
+                        .filter(f -> f.isDirectory())
+                        .map(File::getAbsolutePath)
```
britzl, line 1054: We did this in a later step. Better do it here.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1118,7 +1112,7 @@ private File compileAndroidResources(String platform, Map<String, Object> merged
         return outputDirectory;
     }
 
-    private Map<String, File> linkAndroidResources(File compiledResourcesDir, String platform, Map<String, Object> mergedAppContext) throws ExtenderException {
```
britzl, line 1121: platform was not used

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1877,26 +1872,57 @@ private List<File> copyAndroidResourceFolders(String platform) throws ExtenderEx
                                                       .collect(Collectors.toList());
     }
 
+
+    // get extra packages (for aapt2) from the 'package' attribute in the AndroidManifest
+    // of the gradle dependencies. only get extra packages from aar dependencies which
+    // have a res folder
+    private List<String> getExtraPackagesFromGradlePackages() throws ExtenderException {
```
britzl, line 1879: The juicy bit of this PR

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1877,26 +1872,57 @@ private List<File> copyAndroidResourceFolders(String platform) throws ExtenderEx
...
+        try {
+            for (File f : gradlePackages) {
+                if(f.getName().endsWith(".aar")) {
+                    File res = new File(f, "res");
+                    File androidManifest = new File(f, "AndroidManifest.xml");
+                    if (res.exists() && androidManifest.exists()) {
+                        String am = FileUtils.readFileToString(androidManifest);
+                        Pattern p = Pattern.compile(".*package=\"(.*?)\".*", Pattern.MULTILINE | Pattern.DOTALL);
+                        Matcher m = p.matcher(am);
+                        if (m.matches()) {
```
britzl, line 1890: Possible improvement would be to actually parse the xml instead of using a regex. Is it worth it?

JCash: I think this is fine for now. If it becomes an issue, we can do it later.

JCash (approved):

JCash:
> This reduces or completely removes the need to specify aaptExtraPackages in ext.manifest.

What happens if you have something specified in `aaptExtraPackages` that is now instead generated?

britzl:
> > This reduces or completely removes the need to specify aaptExtraPackages in ext.manifest.
> 
> What happens if you have something specified in `aaptExtraPackages` that is now instead generated?

It will be added twice to `--extraPackages` which doesn't seem to matter. There's still only one R.java per package generated.

PR #249, opened 2022-11-07
Updated host_platform name to x86_64-macos to match our toolset

JCash:


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -159,7 +159,7 @@ private class ProGuardContext {
 
         // These host names are using the Defold SDK names
         if (os.contains("Mac")) {
-            this.hostPlatform = "x86_64-darwin";
+            this.hostPlatform = "x86_64-macos";
```
JCash, line 162: A variable used in out build.yml

britzl (approved):

PR #250, opened 2022-11-09
Add JAVA_HOME to allowed environment variables

JCash:


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -183,6 +183,7 @@ private class ProGuardContext {
             envContext.put("env.LD_LIBRARY_PATH", "."); // Easier when running a standalone local without such a variable
 
             processExecutor.putEnv("DYNAMO_HOME", sdk.getAbsolutePath());
+            processExecutor.putEnv("JAVA_HOME", System.getenv("JAVA_HOME"));
```
JCash, line 186: With this, we can add the JDK includes in an extension

britzl (approved):

PR #251, opened 2022-11-28
Don't add environment variables that are null to the ProcessExecutor

JCash:


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -183,7 +183,11 @@ private class ProGuardContext {
             envContext.put("env.LD_LIBRARY_PATH", "."); // Easier when running a standalone local without such a variable
 
             processExecutor.putEnv("DYNAMO_HOME", sdk.getAbsolutePath());
-            processExecutor.putEnv("JAVA_HOME", System.getenv("JAVA_HOME"));
+            String java_home = System.getenv("JAVA_HOME");
+            if (java_home != null)
+            {
+                processExecutor.putEnv("JAVA_HOME", java_home);
+            }
```
JCash, line 190: We currently haven't setup the JAVA_HOME on the darwin server.
We'll need it for the Rive extension, since we are building jni stuff, and we need the java sdk (jni.h):

```
defold-rive/ext.manifest
    includes:   ["{{env.JAVA_HOME}}/include/linux", "{{env.JAVA_HOME}}/include"]
```

server/src/main/java/com/defold/extender/ProcessExecutor.java
```diff
@@ -127,7 +133,10 @@ public Void call() throws Exception {
             } else if (e.getCause() instanceof InterruptedException) {
                 throw (InterruptedException)e.getCause();
             } else {
-                throw new ExtenderException(e.getCause().toString());
+                StringWriter sw = new StringWriter();
+                PrintWriter pw = new PrintWriter(sw);
+                e.printStackTrace(pw);
+                throw new ExtenderException(sw.toString());
```
JCash, line 139: This actually prints out the full callstack of the error.
I guess the cause, had another cause and we didn't iterate it?

britzl (approved):

PR #260, opened 2023-03-06
Improve internal merging of app/extension manifests

JCash:
The problem was that the merging/overrides was done in an unintuitive manner.
Now, most of the merging code was moved to the `Extender.loadManifests()` function which is called at the start of the job.
The issue at hand was that we needed to override (and exclude) some engine libraries (for the profiler), but the previous way of merging happened too late.

At later stages, the build steps may use the extension or app context to replace environment variables accordingly.


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1794,45 +1698,101 @@ private File buildWin32Resources(Map<String, Object> mergedAppContext) throws Ex
     }
 
     private void loadManifests() throws IOException, ExtenderException {
+        // Creates the contexts for each stage: extension and app
+        //  extension context: merge(platform, variant, extension, app)
+        //  app context:       merge(platform, variant, extensions, app)
```
JCash, line 1703: This function now does a lot of the heavy lifting, in preparing the individual contexts for extensions and the app.

Here is listed the order of operations in order to create the final context:
* extension context: `merge(platform, variant, extension, app)`
* app context:           `merge(platform, variant, extensions, app)`

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -368,15 +389,153 @@ static private String stripMergeKey(String name)
...
+    //  * Same class
+    //  * Uses the public fields
+    //  * Merges Lists and Maps
+    //  * Maps can override values by using the "_replace" name on variables
+    // Merges other on top of dst:
+    //  dst = {"key": valueA, "list" : ListA, "keyA": valueA}
+    //  other = {"key": valueB, "list" : ListB, , "keyB": valueB}
+    //  -> dst: {"key": valueB, "list" : merge(ListA, ListB), , "keyA": valueA, , "keyB": valueB}
+    //@SuppressWarnings("unchecked")
+    public static <T extends Object> void mergeObjects(T dst, T other) throws ExtenderException {
```
JCash, line 459: Helper function to merge two instances of the same type.
E.g. we use it to merge properties of two "PlatformConfig"

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -368,15 +389,153 @@ static private String stripMergeKey(String name)
...
+        PruneMapping(String targetName, String includeName, String excludeName) {
+            this.targetName = targetName;
+            this.includeName = includeName;
+            this.excludeName = excludeName;
+        }
+    };
+
     // Copies the original context, and appends the extra context's elements, if the keys and types are valid
-    static public Map<String, Object> mergeContexts(Map<String, Object> originalContext, Map<String, Object> extensionContext) throws ExtenderException {
+    static public Map<String, Object> mergeContexts(Map<String, Object> a, Map<String, Object> b) throws ExtenderException {
```
JCash, line 508: The "mergeContexts" now also has the responsibility to prune the content when doing the merge.

Previously, this was spread out all over the Extender.java.

This is essentially the actual bug fix, pruning content as we merge each step.

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -421,21 +580,13 @@ static public Map<String, Object> mergeContexts(Map<String, Object> originalCont
...
-            } else if (v instanceof List) {
-                v = new ArrayList<String>();
-            }
-            out.put(k, v);
-        }
-        return out;
+    public static PlatformConfig createPlatformConfig(AppManifestPlatformConfig appconfig) throws ExtenderException {
+        PlatformConfig config = new PlatformConfig();
+        config.context = mergeMaps(config.context, appconfig.context);
+        return config;
```
JCash, line 589: Since we operate on PlatformConfig, we needed an easy convert function.

server/src/main/java/com/defold/extender/TemplateExecutor.java
```diff
@@ -16,7 +16,12 @@ public class TemplateExecutor {
 
     public String execute(String template, Map<String, Object> context) {
         try {
-        	return Mustache.compiler().compile(template).execute(context);
+        	String result = Mustache.compiler().compile(template).execute(context);
+            while (!result.equals(template)) {
+                template = result;
+                result = Mustache.compiler().compile(template).execute(context);
+            }
```
JCash, line 23: A problem we previously had was that if one variable contained another expression, we couldn't resolve that:

```
env:
  a: "foo"
  b: "{{env.a}}"
  c: "{{env.b}}"
```

server/src/main/java/com/defold/extender/TemplateExecutor.java
```diff
@@ -28,7 +33,7 @@ public List<String> execute(List<String> templates, Map<String, Object> context)
     	List<String> out = new ArrayList<>();
     	for (String template : templates) {
         	try {
-    			out.add(Mustache.compiler().compile(template).execute(context));
+    			out.add(this.execute(template, context));
```
JCash, line 36: Use the internal function that resolves recursively.

server/src/test/java/com/defold/extender/ExtenderTest.java
```diff
@@ -355,6 +356,60 @@ public void testAppManifestContext() throws IOException, ExtenderException {
         assertEquals( expectedItems, context.get("flags") );
     }
 
+    @Test
+    public void testMergedContexts() throws IOException, ExtenderException {
```
JCash, line 360: Unit test for my actual use case (removing profiler libraries)

server/src/main/java/com/defold/extender/services/DefoldSdkService.java
```diff
@@ -209,6 +212,10 @@ private void deleteCachedSdk(Path path) {
     @PreDestroy
     @SuppressWarnings("unused")
     public void destroy() {
+        if (!this.cacheClearOnExit) {
+            LOGGER.info("Skipping cleanup of SDK cache");
+            return;
+        }
```
JCash, line 218: By not deleting the cache, the local tests should go faster, by not having to download the sdk's all the time.

britzl (approved):

PR #275, opened 2023-04-14
Fix for supporting different types of pruning patterns

JCash:


server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -26,14 +27,26 @@
 
 public class ExtenderUtil
 {
+    static Pattern createPattern(String expression) {
+        Pattern p = null;
+        try {
+            p = Pattern.compile(expression);
```
JCash, line 33: For some cases, e.g. the pruning pattern is a good regex that we'll use later (e.g. "*.jar")

But in this case, I needed to also support a string like "{{env.SDK}}/path/to/lib.a", but these are reserved characters.

This issue popped up now with the latest refactoring, and this was a case I missed before.

JCash: I am not sure this is the best solution, but it seems to work for now at least.

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -26,14 +27,26 @@
...
+        } catch (PatternSyntaxException e) {
+            p = Pattern.compile(expression, Pattern.LITERAL);
+        }
+        if (p == null) {
+            p = Pattern.compile(expression); // make it throw again
```
britzl, line 38: Can't you throw(e) instead?

britzl (approved):

PR #284, opened 2023-04-21
Add support for compiling plain c

### LINKED ISSUE #272: [Cocoapods] Support compiling plain C
I tried to build Sentry pod, but I got compiler's error regarding C sources
```
clang++ -c -arch arm64 -target arm-apple-darwin19 -m64 -O2 -g -stdlib=libc++ -miphoneos-version-min=11.0 -isysroot <redacted>/platformsdk/iPhoneOS16.2.sdk -nostdinc++ -isystem <redacted>/platformsdk/iPhoneOS16.2.sdk/usr/include/c++/v1  -DDLIB_LOG_DOMAIN="SENTRY" -DDDF_EXPOSE_DESCRIPTORS -DDM_PLATFORM_IOS -DDM_DEBUG  -fno-exceptions -fvisibility=hidden -fobjc-arc -Iupload/ -ICocoaPodsService/Pods/Sentry/  -F/var/folders/nq/zlqy380j2fb1xwwkry71j9jw0000gn/T/job11030230481155200560/CocoaPodsService/frameworks/lib/arm64-ios  -I<redacted>/sdk/b1f3977fbad9230bbf048e67e779d1e3c86f2cf6/defoldsdk//include -I<redacted>/sdk/b1f3977fbad9230bbf048e67e779d1e3c86f2cf6/defoldsdk//sdk/include -I<redacted>/sdk/b1f3977fbad9230bbf048e67e779d1e3c86f2cf6/defoldsdk//ext/include   CocoaPodsService/Pods/Sentry/Sources/SentryCrash/Recording/Tools/SentryCrashStackCursor_MachineContext.c -obuild/SentryCrashStackCursor_MachineContext.c_3.o
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
In file included from CocoaPodsService/Pods/Sentry/Sources/SentryCrash/Recording/Tools/SentryCrashStackCursor_MachineContext.c:28:
CocoaPodsService/Pods/Sentry/Sources/SentryCrash/Recording/Tools/SentryCrashMemory.h:69:26: error: expected ')'
    const void *restrict const src, void *restrict const dst, int byteCount);
```
As I understood `restrict` keyword used only in C.
Podfile:
```
platform :ios, '11.0'

pod 'Sentry', :git => 'https://github.com/getsentry/sentry-cocoa.git', :tag => '7.31.5'
```

britzl:
Fixes #272

### DIFF
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -680,28 +680,7 @@ class Extender {
         List<String> objs = new ArrayList<>();
         List<String> commands = new ArrayList<>();
 
-        // build lists of flags and defines per supported language
-        List<String> defines = new ArrayList<>(pod.defines);
-        List<String> flagsC = new ArrayList<>(pod.flags);
-        List<String> flagsCpp = new ArrayList<>(pod.flags);
-        List<String> flagsObjC = new ArrayList<>(pod.flags);
-        List<String> flagsObjCpp = new ArrayList<>(pod.flags);
-        if (platform.contains("ios")) {
-            defines.addAll(pod.ios_defines);
-            flagsC.addAll(pod.ios_flags);
-            flagsCpp.addAll(pod.ios_flags);
-            flagsObjC.addAll(pod.ios_flags);
-            flagsObjCpp.addAll(pod.ios_flags);
-        }
-        else if (platform.contains("osx")) {
-            defines.addAll(pod.osx_defines);
-            flagsC.addAll(pod.osx_flags);
-            flagsCpp.addAll(pod.osx_flags);
-            flagsObjC.addAll(pod.osx_flags);
-            flagsObjCpp.addAll(pod.osx_flags);
-        }
-
-        // TEMP FIX! remove hardcoded flags from context (should be removed from build.yaml perhaps?)
+        // remove hardcoded flags from context
         Map<String, Object> trimmedContext = ExtenderUtil.mergeContexts(manifestContext, new HashMap<>());
         removeFlag("-fno-exceptions", (List<String>)trimmedContext.get("flags"));
         removeFlag("-stdlib=libc++", (List<String>)trimmedContext.get("flags"));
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -680,59 +680,101 @@ private List<String> compilePodSourceFiles(PodSpec pod, Map<String, Object> mani
...
             flagsCpp.addAll(pod.osx_flags);
             flagsObjC.addAll(pod.osx_flags);
-            flagsC.addAll(pod.osx_flags);
-            defines.addAll(pod.osx_defines);
+            flagsObjCpp.addAll(pod.osx_flags);
         }
-        removeFlag("-std=c++", flagsObjC);
-        removeFlag("-std=c++", flagsC);
+
+        // TEMP FIX! remove hardcoded flags from context (should be removed from build.yaml perhaps?)
```
britzl, line 704: Or maybe this isn't a temp fix? We probably only want to let pods toggle use of exceptions?

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -680,59 +680,66 @@ private List<String> compilePodSourceFiles(PodSpec pod, Map<String, Object> mani
...
-        podContextObjC.put("flags", flagsObjC);
-        podContextObjC.put("defines", defines);
-        podContextC.put("flags", flagsC);
-        podContextC.put("defines", defines);
-        Map mergedContextWithPodsForCpp = ExtenderUtil.mergeContexts(manifestContext, podContextCpp);
-        Map mergedContextWithPodsForObjC = ExtenderUtil.mergeContexts(manifestContext, podContextObjC);
-        Map mergedContextWithPodsForC = ExtenderUtil.mergeContexts(manifestContext, podContextC);
+        Map<String, Object> podContextObjCpp = new HashMap<>();
+
+        if (platform.contains("ios")) {
```
JCash, line 695: What are examples of these flags?
Is it a complete set?
Can we use them as-is?
Do we really need our compile flags?
Perhaps, if it's just paths we need, those paths should be better specified in out `build_input.yml`?

server/src/main/java/com/defold/extender/services/CocoaPodsService.java
```diff
@@ -115,25 +196,14 @@ public class PodSpec {
         public Set<File> includePaths = new HashSet<>();
```
JCash, line 196: We don't want to use Set here, as it's unordered, and for compile flags, it needs to be controlled.
E.g. you don't want to switch order of include paths, as that might get you a different header file each compile run.

JCash (changes requested):
Overall goo imo.

* Need to move away from Set as a container for flags (that'sa why we use Lists all other places)
* I'm concerned about our hardcoded manipulation of our compile commands from `build_input.yml`.

server/src/main/java/com/defold/extender/services/CocoaPodsService.java
britzl: Good point! I've changed to a LinkedHashSet (ie a set with a predictable iteration order based on insertion order)

server/src/main/java/com/defold/extender/services/CocoaPodsService.java
```diff
@@ -115,25 +196,14 @@ public class PodSpec {
...
-        public Set<String> flags = new HashSet<>();
-        public Set<String> ios_flags = new HashSet<>();
-        public Set<String> osx_flags = new HashSet<>();
-        public Set<String> linkflags = new HashSet<>();
-        public Set<String> ios_linkflags = new HashSet<>();
-        public Set<String> osx_linkflags = new HashSet<>();
+
+        public PlatformAndLanguageSet flags = new PlatformAndLanguageSet();
+        public PlatformSet defines = new PlatformSet();
+        public PlatformSet linkflags = new PlatformSet();
```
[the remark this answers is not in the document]
britzl: Changed to LinkedHashSet

server/src/main/java/com/defold/extender/Extender.java
britzl: I think you are right. The flags we get from the podspec should be a complete set so there's not really any need to merge them.

JCash (approved):

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -680,59 +680,66 @@ private List<String> compilePodSourceFiles(PodSpec pod, Map<String, Object> mani
...
-            flagsObjC.addAll(pod.osx_flags);
-            flagsC.addAll(pod.osx_flags);
-            defines.addAll(pod.osx_defines);
-        }
-        removeFlag("-std=c++", flagsObjC);
-        removeFlag("-std=c++", flagsC);
+        // remove hardcoded flags from context
+        Map<String, Object> trimmedContext = ExtenderUtil.mergeContexts(manifestContext, new HashMap<>());
+        removeFlag("-fno-exceptions", (List<String>)trimmedContext.get("flags"));
+        removeFlag("-stdlib=libc++", (List<String>)trimmedContext.get("flags"));
```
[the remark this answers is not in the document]
britzl: I agree. This will be the next step.

PR #293, opened 2023-05-17
Add resources from cocoapods to build output

### LINKED ISSUE #292: Add framework resources to build output
Some frameworks (resolved by Cocoapods) contains [resources](https://guides.cocoapods.org/syntax/podspec.html#resources) that should be copied into the final application. Make sure we detect these when resolving cocoapod dependencies and add them to the build output that we send back to the client.

britzl:
Fixes #292

### DIFF
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -770,6 +770,25 @@ class Extender {
             }
         }
 
+        LOGGER.info("buildPods - adding framework resource to build output");
+        File resourcesBuildDir = new File(buildDirectory, "resources");
+        resourcesBuildDir.mkdir();
+        List<String> resources = resolvedPods.getAllPodResources(platform);
+        for (String resource : resources) {
+            File resourceFile = new File(resource);
+            if (resourceFile.isFile()) {
+                File resourceDestFile = new File(resourcesBuildDir, resourceFile.getName());
+                Files.copy(resourceFile.toPath(), resourceDestFile.toPath());
+                outputFiles.add(resourceDestFile);
+            }
+            else {
+                File resourceDestDir = new File(resourcesBuildDir, resourceFile.getName());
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -770,6 +770,25 @@ private List<File> buildPods() throws IOException, InterruptedException, Extende
             }
         }
 
+        LOGGER.info("buildPods - adding framework resource to build output");
+        File resourcesBuildDir = new File(buildDirectory, "resources");
+        resourcesBuildDir.mkdir();
+        List<String> resources = resolvedPods.getAllPodResources(platform);
+        for (String resource : resources) {
+            File resourceFile = new File(resource);
+            if (resourceFile.isFile()) {
```
britzl, line 779: Resources can according to the pod spec be either individual resources or .bundle packages (ie folders)

server/src/main/java/com/defold/extender/services/CocoaPodsService.java
```diff
@@ -51,51 +51,42 @@ public class ResolvedPods {
...
-                }
-                else if (platform.contains("osx")) {
-                    flags.addAll(pod.linkflags.osx);
-                }
+                flags.addAll(pod.linkflags.get(platform));
             }
             return new ArrayList<String>(flags);
         }
 
+        public List<String> getAllPodResources(String platform) {
```
britzl, line 67: Get all pod resources. Each resource has a path relative to the pod dir. Return absolute paths.

server/src/main/java/com/defold/extender/services/CocoaPodsService.java
```diff
@@ -177,6 +168,16 @@ public void addAll(List<String> values) {
             }
         }
 
+        public Set<String> get(String platform) {
```
britzl, line 171: Helper to avoid repeating code above

JCash (approved):

PR #295, opened 2023-05-17
Handle static libraries in frameworks

### LINKED ISSUE #294: Handle static libraries in xcframeworks
Multi-arch frameworks (ie xcframework) can contain both platform specific frameworks (ie .framework folders) and static libraries (.a files). We are currently expecting all xcframeworks to only include .frameworks.

britzl:
Fixes #294

### DIFF
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -776,13 +776,13 @@ class Extender {
 
         List<String> paths = getFrameworkPaths(resolvedPods.frameworksDir);
         for (String path : paths) {
-            File[] frameworkDirs = new File(path).listFiles();
-            for (File frameworkDir : frameworkDirs) {
-                if (FrameworkUtil.isDynamicallyLinked(frameworkDir)) {
+            File[] frameworks = new File(path).listFiles();
+            for (File framework : frameworks) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -776,13 +776,13 @@ private List<File> buildPods() throws IOException, InterruptedException, Extende
 
         List<String> paths = getFrameworkPaths(resolvedPods.frameworksDir);
         for (String path : paths) {
-            File[] frameworkDirs = new File(path).listFiles();
```
britzl, line 779: This is nothing more than renaming of variables since a framework can be both a .framework package and a .a static library

server/src/main/java/com/defold/extender/services/CocoaPodsService.java
```diff
@@ -390,7 +390,19 @@ private List<File> findPodIncludePaths(PodSpec pod, String pattern) {
         }
 
         return new ArrayList<File>(includePaths);
-    }    
+    }
+
+    private void copyPodFrameworksFromArchitectureDir(File architectureDir, File toDir) throws IOException {
```
britzl, line 395: Helper to simplify the code in copyPodFrameworks() below

PR #329, opened 2023-11-29
Added config for standalone-dev extender

britzl:


server/src/main/resources/application-local.yaml
```diff
@@ -0,0 +1,15 @@
...
+        enabled: true
+        type: LOCAL
+        local.basedir: /var/extender/cache/data
+        #type: S3
+        #s3.bucket: defold-extender-cache-dev
+
+    remote-builder:
+        enabled: false
+        # This points to the host machine from inside docker
+        url: http://host.docker.internal:9010
```
britzl, line 15: Previously application-dev.yml

PR #331, opened 2023-12-19
Don't log about debug variables when they're not used

JCash:


server/src/main/java/com/defold/extender/ExtenderController.java
```diff
@@ -365,7 +365,10 @@ public void buildEngineAsync(HttpServletRequest _request,
                 }
             }
             else {
-                LOGGER.info("Keeping job folder due to debug flags");
+                if (!deleteDirectory)
+                {
+                    LOGGER.info("Keeping job folder due to debug flags");
+                }
```
JCash, line 371: Since it's an asynchronous operation, I think keeping the job folder is implied.

Also, the message was about the debug variables, and behavior should reflect that.

britzl (approved):

PR #335, opened 2024-01-11
Use long type for handling request body size.

### LINKED ISSUE #200: Handle builds larger than 2gb
The max build job size is parsed to an int which is causing problems on really large build jobs:

Injected here: https://github.com/defold/extender/blob/dev/server/src/main/java/com/defold/extender/ExtenderController.java#L92

Parsed here: https://github.com/defold/extender/blob/dev/server/src/main/java/com/defold/extender/ExtenderController.java#L65

For configurations of `spring.http.multipart.max-request-size` larger than 2048mb we hit the max integer size. We should convert to a long instead and we should use getContentLengthLong() from HttpServletRequest (Java EE 7):

https://docs.oracle.com/javaee/7/api/javax/servlet/ServletRequest.html#getContentLengthLong--

ekharkunov:
Fixes #200.


britzl (approved):
Looks good. Nice improvement!

PR #336, opened 2024-01-11
Check plist array for duplication during manifest merge phase.

### LINKED ISSUE #332: Info.plist should specify CFBundleSupportedPlatforms with an array containing a single platform
Apple sends an error indicating that the info.plist contains two identical keys.
ITMS-90562: Invalid Bundle - Info.plist should specify CFBundleSupportedPlatforms with an array containing a single platform.
In .ipa file info.plist containts:
```
  <key>CFBundleSupportedPlatforms</key>
        <array>
            <string>iPhoneOS</string>
            <string>iPhoneOS</string>
        </array>
```
The application contains a extension that has its own info.plist file and an CFBundleSupportedPlatforms field and value iPhoneOS.
And in the basic info.list there is also this key with the same value.

ekharkunov:
Fixes #332.


AGulev:
Great improvement, special thx for adding test for it!

PR #349, opened 2024-03-04
Don't move empty folders to platformsdk folder. 

### LINKED ISSUE #348: setup-standalone-server.sh leaves empty folders for missing packages
If a package is missing from DM_PACKAGES_URL, download_package() will leave an empty folder in extender/platformsdk/ when running rebuild-and-run-standalone-local.sh

This has two consequences:
1. A visual inspection of extender/platformsdk/ in Finder suggests the item has been copied because the SDK folder is present
2. If the packaged sdk is added to the DM_PACKAGES_URL location and rebuild-and-run-standalone-local.sh is run again, the folder is detected and the download is skipped because it appears to already be installed.

dri-richard:
Fixes #348 
Checks if the download folder is empty, and doesn't copy it to the platformsdk folder if so.

### DIFF
--- a/server/scripts/standalone/setup-standalone-server.sh
+++ b/server/scripts/standalone/setup-standalone-server.sh
@@ -58,11 +58,14 @@ function download_package() {
 		local folder=`(cd ${TMP_DOWNLOAD_DIR} && ls)`
 		echo "[setup] Found folder" ${folder}
 
+		if [[ -n ${folder} ]]; then
+			mv ${TMP_DOWNLOAD_DIR}/${folder} ${PLATFORMSDK_DIR}/${out_package_name}
+			echo "[setup] Installed" ${PLATFORMSDK_DIR}/${package_name}
+		else
+			echo "[setup] Failed to install" ${package_name}
+		fi
 
-		mv ${TMP_DOWNLOAD_DIR}/${folder} ${PLATFORMSDK_DIR}/${out_package_name}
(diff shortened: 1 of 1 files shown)


server/scripts/standalone/setup-standalone-server.sh
```diff
@@ -58,11 +58,14 @@ function download_package() {
 		local folder=`(cd ${TMP_DOWNLOAD_DIR} && ls)`
 		echo "[setup] Found folder" ${folder}
 
+		if [[ -n ${folder} ]]; then
+			mv ${TMP_DOWNLOAD_DIR}/${folder} ${PLATFORMSDK_DIR}/${out_package_name}
+			echo "[setup] Installed" ${PLATFORMSDK_DIR}/${package_name}
+		else
+			echo "[setup] Failed to install" ${package_name}
```
britzl, line 65: Would it make sense to `exit 1` here?

britzl (commented):
Should we exit the script if a package download fails?

server/scripts/standalone/setup-standalone-server.sh
dri-richard: Possibly. It would've helped me, but for experienced users who only want (for example) Xcode 14.2 OR Xcode 15.0, exiting might be frustrating as they would need to download & package unwanted tools/sdks.

britzl: Good point. Let's leave it as-is

PR #356, opened 2024-03-15
Fix the keep job directory flag

britzl:


server/src/main/java/com/defold/extender/AsyncBuilder.java
```diff
@@ -48,7 +48,7 @@ public AsyncBuilder(DefoldSdkService defoldSdkService,
         this.gradleService = gradleService;
         this.cocoaPodsService = cocoaPodsService;
         this.jobResultLocation = new File(jobResultLocation);
-        this.keepJobDirectory = System.getenv("DM_DEBUG_KEEP_JOB_FOLDER") != null && System.getenv("DM_DEBUG_JOB_FOLDER") == null;
+        this.keepJobDirectory = System.getenv("DM_DEBUG_KEEP_JOB_FOLDER") != null || System.getenv("DM_DEBUG_JOB_FOLDER") != null;
```
britzl, line 51: Keep the job directory either if the DM_DEBUG_JOB_FOLDER env var is set (ie a custom job directory is used) or if the DM_DEBUG_KEEP_JOB_FOLDER env var is set

JCash (approved):

PR #376, opened 2024-05-06
Unescape gcc defines before use

### LINKED ISSUE #369: Error building with Yandex Ads 7.0.1 on iOS
Here is the sample project https://github.com/Lerg/extension-yandexads/tree/7.0.1

It produces an error
```
duplicate interface definition for class 'AMAArrayIterator'
@interface AMAArrayIterator : NSObject <AMAResettableIterable>
```

If only `pod 'YandexMobileAds', '7.0.1'` is specified in the Podfile, another error appears
```
The abstract target Pods is not inherited by a concrete target, so the following dependencies won't make it into any targets in your project:
    - YandexMobileAds (= 7.0.1)
```

This issue on the Yandex Ads github:
https://github.com/yandexmobile/yandex-ads-sdk-ios/issues/216

britzl:
The `GCC_PREPROCESSOR_DEFINITIONS` of a Podspec may contain escaped strings. These need to be unescaped before use. Example:

https://github.com/CocoaPods/Specs/blob/master/Specs/1/7/a/AppMetricaPlatform/5.3.1/AppMetricaPlatform.podspec.json#L23

Fixes #369


britzl:
Note: Requires this change to build_input.yml to actually be useful: https://github.com/defold/defold/pull/8888

PR #386, opened 2024-06-03
Add -std=c++11 to flags based on libraries used

### LINKED ISSUE #369: Error building with Yandex Ads 7.0.1 on iOS

britzl:
Fixes #369


AGulev:
it looks good

AGulev (approved):

PR #392, opened 2024-06-12
Split docker

### LINKED ISSUE #252: Add Docker flags to disable specific platforms
It would be nice to have a way to exclude specific platforms' setups in the Dockerfile via environment variables so that users don't need to mess with the build process if they don't have SDKs for every platform. I propose adding `EXCLUDE_<PLATFORM>` build arguments for each supported platform and checking for its existence in every `RUN` command that requires that SDK. It would look something like:

```sh
# build.sh
...
if [ "${EXCLUDE_WINDOWS}" != "" ]; then
	ENV="${ENV} --build-arg EXCLUDE_WINDOWS"
fi
...
```

```dockerfile
# docker-base/Dockerfile
...

#
# Windows
#

ARG EXCLUDE_WINDOWS

...

RUN if [ -z "$EXCLUDE_WINDOWS" ] ; then \
  echo "WIN32 2019 SDK" && \
  mkdir -p ${PLATFORMSDK_WIN32}/MicrosoftVisualStudio2019 && \
  wget -q -O - ${DM_PACKAGES_URL}/Microsoft-Visual-Studio-2019-${WINDOWS_MSVC_2019_VERSION}.tar.gz | tar xz -C ${PLATFORMSDK_WIN32}/MicrosoftVisualStudio2019 \
  ; fi

RUN if [ -z "$EXCLUDE_WINDOWS" ] ; then \
  echo "WIN32 10 SDK" && \
  mkdir -p ${PLATFORMSDK_WIN32}/WindowsKits && \
  wget -q -O - ${DM_PACKAGES_URL}/WindowsKits-${WINDOWS_SDK_10_VERSION}.tar.gz | tar xz -C ${PLATFORMSDK_WIN32}/WindowsKits \
  ; fi

# You get the idea
...
```
With that setup, the user doesn't need to provide Windows SDKs if they build with:
```sh
EXCLUDE_WINDOWS=1 ./server/scripts/build.sh
```

I did something very similar in my Rust fork, but that used a whitelist instead of blacklist which would break existing workflows.

### LINKED ISSUE #289: Readd support for metrics
We want to hook up our metrics to a server again.

### LINKED ISSUE #339: Investigate option to use something other than AWS macOS instances

### LINKED ISSUE #413: Create container for NSDK 18.3.2

### LINKED ISSUE #414: Create container for PS5 9.000

ekharkunov:
* Split docker image into separate images per environment (platform+toolchain/sdk version)
* Add docker compose settings to run splitted version locally
* Separate docker image and application code that runs in pre-configured docker environment.
* Rework how remote builder can be configured. Now it set as dictionary of `<platform>-<sdk_version>: <builder_url>`
* Upgrade spring boot to latest 3.x version.
* Cleanup some unused imports from code.
* Upgrade other Java dependencies.
* Add Github workflows to automate build and push Docker images with environments to private registries.
* Add Github workflow to automate build and push jars to Maven repository.
* Template files that used during engine build (for example, **build.gradle.template**) moved inside jar to simplify deployment.
* Separate gradle service class into two implementation: fake and real. Real used only if `extender.gradle.enabled` property set to `true`. Real gradle service used only for android builder, for others - used fake service.
* ~~Add send metrics to Stackdriver.~~ Use Influx adapter to send metrics to custom metrics backend.
* Fix metrics units that used for some timers (used SECONDS when value passed in MILLISECONDS).
* Add new documentation with overview of all changes.
* Add service to control VM instances if service runs in at GCP cloud. Service track how much time instance doesn't receive any build workload and suspend it. When new build request arrived and remote builder is suspended - service resume it.
* Add `/health_report` endpoint to collect information about remote builder status.
* Add `/maintanance_mode` endpoint to ensure that all remote builder is running before make any deployment.
* Add implementation to store caches in google bucket.
* Add implementation to work with google bucket files as Spring boot resources.
* Add dockerfiles for latest consoles sdks.
* Add a couple of metrics tot track cache operations.
* Fix metrics types.

More details can be found in **server/README.md**.

Fixes #252 
Fixes #289 
Fixes #339 
Fixes #413 
Fixes #414


britzl (approved):
Massive amount of changes and it is a bit hard to review all of it thoroughly. Looks good though as far as I can tell

PR #407, opened 2024-08-09
Health report

ekharkunov:
* Added special service which collects all information about remote builder status, accumulate per platform and return as JSON.
* Added status page ready-to-deploy on Github pages.

Status page example:
Screenshot 2024-08-09 at 11 42 03


britzl (approved):
Looks good!

PR #409, opened 2024-08-12
Add connection timeout in HealthReporterService. Fix Nan in status page.

ekharkunov:
* Fix Nan in index.html (wrong calculation for total operational builder)
* Add connection timeout configuration in HealthReporter. Otherwise in case of suspended VMs we spend lots of time to reach connection timeout and request status of next builder.

### DIFF
--- a/server/html_data/index.html
+++ b/server/html_data/index.html
@@ -88,7 +88,7 @@
         top_block.appendTo(parent_block);
     }
     function handleResponse(response, rootId) {
-        var totalFullyOperatedPlatforms, totalUnreachablePlatforms = 0;
+        var totalFullyOperatedPlatforms = 0, totalUnreachablePlatforms = 0;
         var parent_block_1 = $(rootId).find("#list-1");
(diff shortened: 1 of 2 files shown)


server/html_data/index.html
```diff
@@ -88,7 +88,7 @@ <h3 id="statusTitle" class="panel-title">
         top_block.appendTo(parent_block);
     }
     function handleResponse(response, rootId) {
-        var totalFullyOperatedPlatforms, totalUnreachablePlatforms = 0;
+        var totalFullyOperatedPlatforms = 0, totalUnreachablePlatforms = 0;
```
britzl, line 91: Might be better to have them on individual lines. The initial problem would have been more easily avoided then.

britzl (approved):

server/html_data/index.html
ekharkunov: Ok. Will make as part of other changes

PR #467, opened 2024-09-25
Update documentation and scripts regarding local Extender

ekharkunov:
* Updated scripts to run local Extender without Docker images
* Updated documentation.

### Technical details
I've added separate .env files to handle environment variables across shell scripts and IDEs (at least VSCode). Files can be following:
* .env - common env for every platform
* macos.env - MacOS specific variables
* user.env - generated file that contains user-specific variable (most of them should be calculated depends on user host machine (fodler-specific pathes))


README.md
```diff
@@ -11,60 +11,74 @@ Extender is a build server that builds native extensions of the Defold engine. T
...
+    ```json
+        {
+            "type": "java",
+            "name": "Spring Boot-ExtenderApplication<server>",
+            "request": "launch",
+            "cwd": "${workspaceFolder}",
+            "mainClass": "com.defold.extender.ExtenderApplication",
+            "projectName": "server",
+            "args": "--spring.config.location=classpath:./,file:${workspaceFolder}/server/configs/ --extender.sdk.location=${workspaceFolder}/server/app/sdk --spring.profiles.active=standalone-dev",
+            "envFile": ["${workspaceFolder}/server/envs/.env", "${workspaceFolder}/server/envs/user.env", "${workspaceFolder}/server/envs/macos.env"]
```
JCash, line 81: I guess this part could be an actual .vscode project?

JCash: Also, it there's a way to setup the dependencies in the project as well, it'd be even easier to use 🤔 

Seems there might exist something like that?
https://docs.lextudio.com/blog/different-ways-to-create-vscode-extension-dependencies/#extension-dependencies-in-configuration

PR #489, opened 2024-10-01
Fix issue with common-io exception

ekharkunov:
* Fix issue with common-io exception
* Fix running docker compose when no DYNAMO_HOME is set.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -60,7 +60,6 @@ dependencies {
     implementation('com.samskivert:jmustache:1.16')
     implementation('commons-chain:commons-chain:1.2')
     implementation('org.apache.commons:commons-fileupload2-jakarta:2.0.0-M1')
-    implementation('commons-io:commons-io:2.12.0')
     implementation('org.apache.commons:commons-text:1.12.+')
     implementation('org.apache.commons:commons-compress:1.27.+')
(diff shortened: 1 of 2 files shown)


server/build.gradle
```diff
@@ -60,7 +60,6 @@ dependencies {
     implementation('com.samskivert:jmustache:1.16')
     implementation('commons-chain:commons-chain:1.2')
     implementation('org.apache.commons:commons-fileupload2-jakarta:2.0.0-M1')
-    implementation('commons-io:commons-io:2.12.0')
```
ekharkunov, line 63: Explicitly pin version leads to mismatch in transitive dependencies. `org.apache.commons:commons-compress` requires at least commons-io:commons-io:2.16.x

server/docker/common-services.yml
```diff
@@ -14,7 +14,7 @@ services:
       - ./../app/:/app/:ro
       - ./../app/:/etc/extender/apps/:ro
       - ./../configs:/etc/defold/extender:ro
-      - ${DYNAMO_HOME:-dummy:/dynamo_home}
```
ekharkunov, line 17: In such configuration if DYNAMO_HOME is not set - it creates a emptyr folder `dummy` and mount it.

JCash (approved):

PR #493, opened 2024-10-02
Update documentation

### LINKED ISSUE #487: The user.env doesn't handle paths with spaces.
When I ran `./server/scripts/standalone/setup-standalone-env.sh ` I got:

```
Load user env ...
/Users/mawe/work/extender/server/scripts/standalone/../../envs/user.env: line 16: Studio: command not found
```

It is due to the fact that the PATH contains paths with spaces in them and it isn't wrapped in ""

ekharkunov:
* Update documentation about running local Extender.
* Fix issue with spaces in PATH variable.
* Use environment variables to define sdk/toolchain package version.
* For standalone version don't cleanup defold sdk cache on exit.
* Add document with metrics description.

Fixes #487


server/build.gradle
```diff
@@ -84,11 +84,20 @@ task createVersionFile() {
     def sha = System.getenv("COMMIT_SHORT_SHA")
     if (sha == null) {
         def outputStream = new ByteArrayOutputStream()
-        exec {
-            commandLine "git", "rev-parse", "--short", "HEAD"
-            standardOutput = outputStream
+        try {
```
ekharkunov, line 87: In case if project was downloaded from github as archive and doesn't have .git folder.

PR #502, opened 2024-10-08
Added DotNet support for Windows target platform

JCash:


server/docker/Dockerfile.winsdk.2022-env
```diff
@@ -86,4 +86,11 @@ RUN \
...
-  cp -v ./gl/*.* ./GL/
\ No newline at end of file
+  cp -v ./gl/*.* ./GL/
+
+# Since dotnet cannot really cross compile, we need to create a "lib" shim for "wine lib.exe"
+# As long as it's in the path, it will be picked up
+RUN \
+  echo '#!/usr/bin/env bash' > /usr/bin/lib && \
+  echo 'wine ${WINDOWS_VCINSTALLDIR_2022}/bin/Hostx64/x64/lib.exe $*' >> /usr/bin/lib && \
+  chmod +x /usr/bin/lib
```
JCash, line 96: It's a hack, but a small price to pay to avoid using an windows machine

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1075,7 +1073,7 @@ private List<File> buildExtensionInternal_CSharp(File manifest, Map<String, Obje
         CSharpBuilder csBuilder = new CSharpBuilder(processExecutor, templateExecutor, context);
         csBuilder.setSdkProject(sdkProject);
         csBuilder.setSourceDirectory(extDir);
-        csBuilder.setOutputDirectory(new File(buildDirectory, "cs"));
+        csBuilder.setOutputDirectory(new File(buildDirectory, extDir.getName()));
```
JCash, line 1076: Each extension should have its own output folder

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
```diff
@@ -91,9 +91,11 @@ public void setSdkProject(File csProject) {
     }
 
     private File writeProject() throws IOException {
+        // The cross compilation (win32) doesn't like absolute unix paths, so we use the path relative to cwd
+        Path relativeOutputDir = sourceDir.toPath().relativize(outputDir.toPath());
```
JCash, line 95: Making sure the path starts with "../../" in order for windows tools to support it.

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
```diff
@@ -177,35 +202,56 @@ private static Path getNativePath(String platform) throws IOException {
...
+        paths.add(getLibName(platform, "Runtime.WorkstationGC"));
+        paths.add(getLibName(platform, "eventpipe-enabled"));
+        paths.add(getLibName(platform, "standalonegc-enabled"));
 
-        linkFlags.add(Paths.get(aotBase.toString(), "libSystem.Native.a").toString());
-        linkFlags.add(Paths.get(aotBase.toString(), "libSystem.IO.Compression.Native.a").toString());
-        linkFlags.add(Paths.get(aotBase.toString(), "libSystem.Globalization.Native.a").toString());
+        String aotSuffix = "";
+        if (isWindows(platform))
+            aotSuffix = ".Aot";
```
JCash, line 227: Yet another weird thing they do, naming things differently between platforms...

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
```diff
@@ -177,35 +189,56 @@ private static Path getNativePath(String platform) throws IOException {
         return Paths.get(NUGET_PACKAGES, String.format("microsoft.netcore.app.runtime.nativeaot.%s/%s/runtimes/%s/native", csplatform, dotnetVersion, csplatform));
     }
 
+    private static ArrayList<String> makePathsAbsolute(String basePath, ArrayList<String> files) {
```
JCash, line 192: Currently still needed to have absolute paths for the libraries on macos, as they ship both static and dynamic libs in the same folder (and clang defaults to choosing the dynamic ones)

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
```diff
@@ -177,35 +189,56 @@ private static Path getNativePath(String platform) throws IOException {
...
+            paths.add(getLibName(platform, "System.Native"));
+            paths.add(getLibName(platform, "stdc++compat"));
+            paths.add(getLibName(platform, "System.Net.Security.Native"));
+            paths.add(getLibName(platform, "System.Security.Cryptography.Native.Apple"));
             linkFlags.add("-licucore");
         }
+        else if (ExtenderUtil.isWindowsTarget(platform))
+        {
+            paths.add(getLibName(platform, "Runtime.VxsortEnabled"));
+            linkFlags.add("-lbcrypt");
```
ekharkunov, line 234: Maybe it can be moved to `build_input.yml`?

ekharkunov (approved):

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
JCash: Yeah, I guess I'm still trying to figure out what's the most convenient way of doing things.
Now that we have the "platform.sdks.json", we can match sdk and server, we may put more things in the server code (a single place)

JCash: Also, those libs dependencies come from DotNet which are solely manged by the server.

PR #529, opened 2024-10-22
Allow setting emscriptenLinkFlags from manifest

Jhonnyg:


ekharkunov (approved):
Please, create a separate task for validation link options and whitelist allowed flags.

ekharkunov:
> > Please, create a separate task for validation link options and whitelist allowed flags.
> 
> Can you elaborate what you mean by this? Do you mean that we should only allow certain emscripten defines for that field?

Yes, only certain options that we alloed to use. Like we do for other fields. For example, https://github.com/defold/defold/blob/dev/share/extender/build_input.yml#L162. And than used here https://github.com/defold/extender/blob/dev/server/src/main/java/com/defold/extender/ExtensionManifestValidator.java#L83-L90

Jhonnyg:
I can add format validation similar to defines, but I don't think we need to add whitelist entries for all the options, there are quite a few of them: https://github.com/emscripten-core/emscripten/blob/main/src/settings.js and keeping our list in sync with theirs will be difficult.

PR #541, opened 2024-10-28
Added ps4-12000 and ps5-10000 to local docker builds

JCash:
Removed old unused emsdk, nssdk, winsdk and ps sdk

### DIFF
--- a/server/configs/application-local-dev-app.yml
+++ b/server/configs/application-local-dev-app.yml
@@ -6,36 +6,21 @@ extender:
                 url: http://android-ndk25:9000
                 instanceId: android-ndk25
                 alwaysOn: true
-            emsdk-2011:
-                url: http://emsdk-2011:9000
-                instanceId: emsdk-2011
-            emsdk-3155:
-                url: http://emsdk-3155:9000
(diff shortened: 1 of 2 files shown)


server/configs/application-local-dev-app.yml
```diff
@@ -6,38 +6,23 @@ extender:
...
-            winsdk-2019:
-                url: http://winsdk-2019:9000
-                instanceId: winsdk-2019
+                instanceId: nssdk-1832
+            ps4-12000:
+                url: http://ps4-12000:9000
+                instanceId: ps4-12000
+            ps5-10000:
+                url: http://ps5-10000:9000
+                instanceId: ps5-10000
```
JCash, line 23: I had forgot to add these, so I could build and test locally

server/configs/application-local-dev-app.yml
```diff
@@ -6,38 +6,23 @@ extender:
                 url: http://android-ndk25:9000
                 instanceId: android-ndk25
                 alwaysOn: true
-            emsdk-2011:
```
JCash, line 9: No need for these locally anymore imho

PR #547, opened 2024-11-08
Added support for building C# for Android

JCash:


server/scripts/standalone/setup-standalone-env.sh
```diff
@@ -111,36 +111,37 @@ function download_zig() {
...
-    # There are no static download links, they're always generated
     # https://dotnet.microsoft.com/en-us/download/dotnet/9.0
+    # https://github.com/dotnet/core/blob/main/release-notes/9.0/install.md
 
-    if [ -d "${DOTNET_ROOT}" ]; then
-        echo "[setup] Removing installed version of dotnet" ${DOTNET_ROOT}
-        rm -rf ${DOTNET_ROOT}
-    fi
+    local version="9.0.100-rc.2.24474.11"
+    if [[ ! -e ${DOTNET_ROOT} ]]; then
```
JCash, line 118: Made sure it only reinstalls if the folder doesn't exist

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -2242,8 +2242,19 @@ private List<File> buildEngine() throws ExtenderException {
...
+            List<String> excludeSymbols = ExtenderUtil.getStringList(mergedAppContext, "excludeSymbols");
+
+            // Not sure what the best way to do this is, but
+
             List<String> symbols = getSortedKeys(manifestConfigs.keySet());
             for (String extensionSymbol : symbols) {
+                if (excludeSymbols.contains(extensionSymbol)) {
+                    LOGGER.info("Skipping extension {} due to excludeSymbols in app manifest", extensionSymbol);
+                    continue;
+                }
```
JCash, line 2256: This helps with disabling an extension, without need to be overly intrusive to the game project (removing dependencies).
Instead, they can use an appmanifest that simply "excludes" an extension.

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
```diff
@@ -197,7 +203,42 @@ private static ArrayList<String> makePathsAbsolute(String basePath, ArrayList<St
...
+        File exportsFile = new File(String.format("%s/defold_cs.exports", buildDir));
+
+        String exportsPattern = "-Wl,--version-script=%s";
+        if (ExtenderUtil.isMacOSTarget(platform) || ExtenderUtil.isIOSTarget(platform))
+        {
+            exportsPattern = "-exported_symbols_list \"%s\"";
+        }
+
+        // HACK: In anticipation of the fix for https://github.com/dotnet/runtime/issues/109341
+        // we have to make sure not all symbols are public, and at the same time respect the engine symbols
```
JCash, line 217: Another wait for bug fixes. But in the meantime this should work for most cases. And should be relatively easy to update if we need to.

server/scripts/standalone/setup-standalone-env.sh
```diff
@@ -111,36 +111,37 @@ function download_zig() {
...
 
-        ${CURL_CMD} -L https://dot.net/v1/dotnet-install.sh --output ./dotnet-install.sh
-        chmod +x ./dotnet-install.sh
+            ${CURL_CMD} -L https://dot.net/v1/dotnet-install.sh --output ./dotnet-install.sh
+            chmod +x ./dotnet-install.sh
 
-        echo "Installing dotnet ..."
-        ./dotnet-install.sh --channel 9.0.1xx --quality preview --install-dir ${DOTNET_ROOT}
+            echo "Installing dotnet ..."
+            ./dotnet-install.sh --version ${9.0.100-rc.2.24474.11} --install-dir ${DOTNET_ROOT}
```
ekharkunov, line 131: I guess misprint here. Should it be `${version}` instead of `${9.0.100-rc.2.24474.11}`?
And here https://github.com/defold/extender/blob/dev/server/scripts/standalone/setup-standalone-env.sh#L153 we also define `version`. Can we use predefined value instead of calculation it every environment update?

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -1379,7 +1379,7 @@ private List<File> linkEngine(List<String> symbols, Map<String, Object> linkCont
         mainContext.put("ext", ImmutableMap.of("symbols", ExtenderUtil.makeUnique(extSymbols)));
 
         String main = templateExecutor.execute(config.main, mainContext);
-        FileUtils.writeStringToFile(maincpp, main);
+        FileUtils.writeStringToFile(maincpp, main, Charset.defaultCharset());
```
ekharkunov, line 1382: 👍

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
```diff
@@ -197,7 +203,42 @@ private static ArrayList<String> makePathsAbsolute(String basePath, ArrayList<St
...
+        String contents = "";
+        contents += "V1.0 { \n";
```
ekharkunov, line 219: Suggestion: create a resource file (the same as we do for podfile.template) and read it instead of building string every time. It can be useful in the future if exports list will become a template.

ekharkunov (changes requested):
Some questions about versioning in setup enviroment script. And suggestion about .Net export list generation.

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
JCash: Yeah, this is still just a hack, until I can get hold of Dotnet 10, where this should be fixed. So I opted out of creating the template file etc.

ekharkunov (approved):

PR #554, opened 2024-11-12
Rework compilation errors handling

### LINKED ISSUE #550: Rework how compilation error is handled
Now when remote builder reports error before send result to client Extender read all compilation output to string and than write to file and to log.

How it should be:
1. **File**: stream data from response entity directly to destination file.
2. **Log**: doesn't put compilation result to log.

Place in code:
* https://github.com/defold/extender/blob/dev/server/src/main/java/com/defold/extender/remote/RemoteEngineBuilder.java#L191
* https://github.com/defold/extender/blob/dev/server/src/main/java/com/defold/extender/remote/RemoteEngineBuilder.java#L199
* (maybe something else)

ekharkunov:
* Compilation error from remote builder now writes directly to result file without pushing it to logs.
* Add `consumeQuite` to make sure that response entity was fully consumed. Add test to check compilation error.
* Add async build variant to integration test configuration.

Fixes #550

### DIFF
--- a/server/src/main/java/com/defold/extender/remote/RemoteEngineBuilder.java
+++ b/server/src/main/java/com/defold/extender/remote/RemoteEngineBuilder.java
@@ -43,6 +43,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
 import java.util.Optional;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
 @Service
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/defold/extender/remote/RemoteEngineBuilder.java
```diff
@@ -188,20 +189,20 @@ public void buildAsync(final RemoteInstanceConfig remoteInstanceConfig,
                     File targetResult = new File(resultDir, BuilderConstants.BUILD_RESULT_FILENAME);
                     Files.move(tmpResult.toPath(), targetResult.toPath(), StandardCopyOption.ATOMIC_MOVE);
                 } else {
-                    String errorLog = EntityUtils.toString(response.getEntity());
-                    LOGGER.error(Markers.COMPILATION_ERROR, String.format("Failed to build source.\n%s", errorLog));
+                    LOGGER.error(Markers.COMPILATION_ERROR, "Failed to build source.");
```
ekharkunov, line 192: I think we don't want to see full build log in logging system. It should be enough to see error at the last step which should be logged in the moment of fail. And than we can collect full work log by traceId.

server/src/test/java/com/defold/extender/IntegrationTest.java
```diff
@@ -90,15 +90,17 @@ public DefoldVersion(String sha1, Version version, String[] platforms)
     private static class TestConfiguration {
         public DefoldVersion version;
         public String platform = "";
+        public boolean async = false;
 
-        public TestConfiguration(DefoldVersion version, String platform) {
+        public TestConfiguration(DefoldVersion version, String platform, boolean async) {
             this.version = version;
             this.platform = platform;
+            this.async = async;
```
ekharkunov, line 98: Add `async` flag to run client build in two modes: sync and async.

britzl (approved):

Issue #393, opened 2024-06-13
[d8] Unsupported usage of main-dex list


ekharkunov:
When compiling Android get following warning
```
Warning in /tmp/job1736659641429178947/build/main_dex_list.txt:
Unsupported usage of main-dex list. The usage of main-dex-list content for the compilation of non-DEX inputs is deprecated. See issue https://issuetracker.google.com/181858113 for context.
```
More details:
https://issuetracker.google.com/181858113

ekharkunov:
Fixed in https://github.com/defold/extender/pull/563

PR #563, opened 2024-11-19
Update Java version to JDK 21.0.5+11

vlaaad:
I just changed the code, do I need to test it? how?

### DIFF
--- a/server/docker/Dockerfile.android.ndk25-env
+++ b/server/docker/Dockerfile.android.ndk25-env
@@ -1,4 +1,4 @@
-FROM europe-west1-docker.pkg.dev/extender-426409/extender-public-registry/extender-android-env:1.3.3
+FROM europe-west1-docker.pkg.dev/extender-426409/extender-public-registry/extender-android-env:1.4.0
 
 #
 # Android SDK/NDK
@@ -98,6 +98,15 @@ RUN --mount=type=secret,id=DM_PACKAGES_URL,required=true \
 # check that aapt installed correctly
     aapt v  && \
     aapt2 version
+
+# download and install R8 utility separatly. It's done in that way because
+# R8/D8 received fixes from time to time which not populated with build tools
+# It's ok that R8.jar downloaded but saved to D8.jar. Can't find D8 as separate utility.
+# R8 contains all classes of D8.
+ENV R8_VERSION=8.7.0-dev
+RUN \
+    wget -O ${ANDROID_SDK_BUILD_TOOLS_PATH_34}/lib/d8.jar https://storage.googleapis.com/r8-releases/raw/${R8_VERSION}/r8.jar && \
+    wget -O ${ANDROID_SDK_BUILD_TOOLS_PATH_33}/lib/d8.jar https://storage.googleapis.com/r8-releases/raw/${R8_VERSION}/r8.jar
 
 # Since dotnet cannot really cross compile, we need to create a "ar" shim for "llvm-ar"
 # As long as it's in the path, it will be picked up
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -2094,7 +2094,9 @@ class Extender {
         try {
             mainList.createNewFile();
             for (String classFile : mainClassNames) {
-                FileUtils.writeStringToFile(mainList, classFile + "\n", Charset.defaultCharset(), true);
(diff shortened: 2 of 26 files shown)


ekharkunov (changes requested):
I'll test it for you.

Please, make changes here:
* https://github.com/defold/extender/blob/dev/.github/workflows/application-build.yml#L40 (and below)
* https://github.com/defold/extender/blob/dev/server/test-data/jars/build.sh#L21 (and below)
* https://github.com/defold/extender/blob/dev/server/test-data/sdk/a/defoldsdk/share/java/build.sh#L2

vlaaad:
I have now added the missed changes.

server/docker/Dockerfile.android.ndk25-env
```diff
@@ -99,6 +99,15 @@ RUN --mount=type=secret,id=DM_PACKAGES_URL,required=true \
     aapt v  && \
     aapt2 version
 
+# download and install R8 utility separatly. It's done in that way because
+# R8/D8 received fixes from time to time which not populated with build tools
+# It's ok that R8.jar downloaded but saved to D8.jar. Can't find D8 as separate utility.
+# R8 contains all classes of D8.
+ENV R8_VERSION=8.7.0-dev
```
ekharkunov, line 106: Install standalone version of R8/D8 utility. Fixed NPE in D8 when function argument doesn't have name (only type).

server/docker/Dockerfile.android.ndk25-env
```diff
@@ -99,6 +99,15 @@ RUN --mount=type=secret,id=DM_PACKAGES_URL,required=true \
...
+RUN \
+    wget -O ${ANDROID_SDK_BUILD_TOOLS_PATH_34}/lib/d8.jar https://storage.googleapis.com/r8-releases/raw/${R8_VERSION}/r8.jar && \
+    wget -O ${ANDROID_SDK_BUILD_TOOLS_PATH_33}/lib/d8.jar https://storage.googleapis.com/r8-releases/raw/${R8_VERSION}/r8.jar
```
ekharkunov, line 109: Updated both because during the build it picks from build tools 33.0.1. It happened because ProcessExecutor doesn't respect PATH variable when start the process.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -2094,7 +2094,9 @@ private File buildMainDexList(List<String> jars) throws ExtenderException {
         try {
             mainList.createNewFile();
             for (String classFile : mainClassNames) {
-                FileUtils.writeStringToFile(mainList, classFile + "\n", Charset.defaultCharset(), true);
+                // create main dex list in form of Proguards rules. Additional info https://github.com/defold/extender/issues/393
+                classFile = classFile.replace("/", ".").replace(".class", "");
+                FileUtils.writeStringToFile(mainList, String.format("-keep class %s { *; }\n", classFile), Charset.defaultCharset(), true);
```
ekharkunov, line 2099: Create main dex list in form of Proguard rules.

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -2114,6 +2116,10 @@ private File[] buildClassesDex(List<String> jars, File mainDexList) throws Exten
         context.put("engineJars", empty_list);
         context.put("mainDexList", mainDexList.getAbsolutePath());
 
+        // replace parameter name because '--main-dex-list' is deprecated and reported as error
+        // we can't change command format for older version of engine so replace parameter here.
+        // Remove when old versions won't be supported.
+        platformConfig.dxCmd = platformConfig.dxCmd.replace("--main-dex-list", "--main-dex-rules");
```
ekharkunov, line 2122: Hacky solution here. `--main-dex-list` is deprecated (see https://github.com/defold/extender/issues/393) and in the latest version it's reported as error. So need to pass `--main-dex-rules` instead. But older engine can't be changed so here we replace argument before pass to template processor.

server/src/main/java/com/defold/extender/Extender.java
vlaaad: Does it lead to increasing the build size? Or is it merely a different syntax?

ekharkunov: Different syntax. Here is nothing happened regarding build shrinking.

ekharkunov (approved):

PR #567, opened 2024-11-22
Fix service state confition.

ekharkunov:
* Fix incorrect condition.
* Check for 'STAGING' state (instance is starting)

### DIFF
--- a/server/src/main/java/com/defold/extender/services/HealthReporterService.java
+++ b/server/src/main/java/com/defold/extender/services/HealthReporterService.java
@@ -64,13 +64,15 @@ public class HealthReporterService {
                 String instanceId = entry.getValue().getInstanceId();
                 // if instance controlled by instanceService and is currently suspended or suspending - mark it as 'operational'
                 if (instanceService != null && instanceService.isInstanceControlled(instanceId)) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/defold/extender/services/HealthReporterService.java
```diff
@@ -64,13 +64,15 @@ public String collectHealthReport(boolean isRemoteBuildEnabled, Map<String, Remo
...
+                    if (instanceService.isInstanceSuspended(instanceId) 
+                    || instanceService.isInstanceSuspending(instanceId)
+                    || instanceService.isInstanceStarting(instanceId)) {
                         updateOperationalStatus(platformOperationalStatus, platform, true);
                         continue;
+                    } else if (!instanceService.isInstanceRunning(instanceId)) {
+                        updateOperationalStatus(platformOperationalStatus, platform, false);
+                        continue;
                     }
-                } else if (!instanceService.isInstanceRunning(instanceId)) {
```
ekharkunov, line 71: Instance should check `isInstanceControlled`. Otherwise we will get error for instance that located outside of GCP.

PR #573, opened 2024-11-30
Use full blob key when touch object in the cache

ekharkunov:


server/src/main/java/com/defold/extender/cache/GCPDataCache.java
```diff
@@ -44,13 +44,14 @@ public boolean exists(String key) {
...
     @Override
     public void touch(String key) {
-        Blob blob = storage.get(this.bucketName, getBlobKey(key));
+        String fullKey = getBlobKey(key);
+        Blob blob = storage.get(this.bucketName, fullKey);
         if (blob != null) {
             try {
                 // metadata can be updated only by copying object to itself
-                blob.copyTo(this.bucketName, key);
+                blob.copyTo(this.bucketName, fullKey);
```
ekharkunov, line 52: Here was an error. File was copied to other location because Extender didn't use full key (with folder prefix).

britzl (approved):
Looks good!

PR #594, opened 2025-01-08
Add support of virtual threads

### LINKED ISSUE #584: Enable virtual threads
Enable "virtual threads" that available in JDK21.

Useful document: https://www.baeldung.com/spring-6-virtual-threads

**Expected outcome**
Less resource should be allocated for thread -> less resources will be waste for threading in Extender -> more resources will be used for building tasks.

ekharkunov:
Reconfigure executor to support virtual threads. Set virtual thread enable flag to false by default.

Fixes #584


server/src/main/resources/application.yml
```diff
@@ -47,6 +47,9 @@ spring:
     # Don't check for templates on startup
     mustache:
         check-template-location: false
+    threads:
+        virtual:
+            enabled: false
```
ekharkunov, line 52: Set default to false to control on deployment and not to break previous behavior (in case of any problem)

britzl (approved):
I'm not familiar with virtual threads in java, but the changes look ok

PR #596, opened 2025-01-09
Dynamic libs

ekharkunov:
Collect from extensions and pack dynamic libraries to build result.
Add tests and test data for new functionality.

Fixes https://github.com/defold/defold/issues/3334


server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -674,6 +674,28 @@ static public List<String> collectFilesByPath(File dir, String re) {
         return result;
     }
 
+    // collect file pathes of all files which name mathes 're'
```
britzl, line 677: Tiny nitpick: It is "paths", not "pathes"

britzl (approved):
Looks good. See note about typo!

AGulev:
It works only when bundle, or when build for the editor as well? (I didn't check )

ekharkunov:
> It works only when bundle, or when build for the editor as well? (I didn't check )

I've checked Macos build of extension-steam from the Editor. It worked.

aglitchman:
Could you please tell me if this new feature works for any version of Defold, i.e. can I already use it on 1.9.6-1.9.7 or do I have to wait for 1.9.8?

ekharkunov:
> Could you please tell me if this new feature works for any version of Defold, i.e. can I already use it on 1.9.6-1.9.7 or do I have to wait for 1.9.8?

You need wait for next Extender's release (after 2.8.2 version). No engine side changes required so it should work for all supported engine versions.

PR #611, opened 2025-01-16
Fix swift compilation

ekharkunov:
Now module map and swift header generates every time except of cases when:
1. Pod spec has "module_map" == "false" or
2. Pod spec has "module_map" == "<path_to_module_map>"

According to https://guides.cocoapods.org/syntax/podspec.html#module_map default value is "true"

Fixes https://github.com/defold/extension-ironsource/issues/36


britzl (approved):
Nice

PR #613, opened 2025-01-16
Readded temporary hack for Dotnet 10 prerelease version

JCash:
This reverts commit 1a21296f953eeceb87ac704c946a429b90aa2930.


ekharkunov (approved):

ekharkunov:
Do we need cleanup all occurrence of `DOTNET_ROOT` in Dockerfile, setup-env scripts?

JCash:
Not sure if we need DOTNET_ROOT at all now tbh. I think the CLI var supercedes it.

PR #662, opened 2025-02-20
Update base Docker image Ubuntu version

### LINKED ISSUE #605: Update base docker image
Because Ubuntu 20.04 end of life is Apr 2025 base docker image need to be updated. Look at Ubuntu 24.04.

ekharkunov:
Fixes #605

### DIFF
--- a/server/docker/Dockerfile.base-env
+++ b/server/docker/Dockerfile.base-env
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
 RUN \
   apt-get update && \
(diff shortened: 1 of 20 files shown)


server/docker/Dockerfile.base-env
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
```
JCash, line 1: I love it!
I've wanted to update that so many times, but the scope was always so large.
With this setup, if feels so much easier to make it all work.
Well done!

JCash (approved):

PR #663, opened 2025-02-21
Support arm64 build

### LINKED ISSUE #660: Add support of arm64-linux
- [x] Update supported platforms in Extender's code
- [x] Update Integration tests (think how to organize testing at different machines)
- [x] Create arm64 linux based Docker images
- [x] Update CI to build arm64-linux docker image

ekharkunov:
Fixes #660

### DIFF
--- a/server/docker/Dockerfile.base-env
+++ b/server/docker/Dockerfile.base-env
@@ -1,5 +1,7 @@
 FROM ubuntu:22.04
 
+ARG TARGETARCH
+
 RUN \
   apt-get update && \
   apt-get update && \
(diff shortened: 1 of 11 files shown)


server/docker/Dockerfile.base-env
```diff
@@ -1,5 +1,7 @@
 FROM ubuntu:22.04
 
+ARG TARGETARCH
```
JCash, line 3: Regarding having a separate arm64 container, I think the biggest benefit is we don't have to fiddle with installing correct cross compile libraries.

As a drawback I wonder what the performance loss is, as I assume this container will still run on a X64 machine.
Regardless, I think this approach is best and easiest, and I don't expect much traffic to this container tbh.

server/src/main/java/com/defold/extender/builders/CSharpBuilder.java
```diff
@@ -122,6 +122,7 @@ private static String convertPlatform(String platform) {
         if (platform.equals("x86_64-win32"))    return "win-x64";
         if (platform.equals("x86-win32"))       return "win-x86";
         if (platform.equals("x86_64-linux"))    return "linux-x64";
+        if (platform.equals("arm64-linux"))     return "linux-arm64";
```
JCash, line 125: 👍

JCash (approved):
lgtm!

server/docker/Dockerfile.base-env
ekharkunov: Here we build a multiarch image (https://docs.docker.com/build/building/multi-platform/). E.g. that we build 2 images (for linux/amd64 and for linux/arm64) which are merged into single image. Than Docker pull image according to host arch or to `--platform` flag.

Arm64 container will be run at arm64 host, so no performance issue will be.

PR #706, opened 2025-04-21
Speedup maintenance mode

### LINKED ISSUE #636: Speedup set maintenance mode
Set maintenance mode are doing by sending requests to every instance one-by-one. Rework it and do in concurrent manner.

ekharkunov:
Fixes #636

### DIFF
--- a/server/src/main/java/com/defold/extender/services/GCPInstanceService.java
+++ b/server/src/main/java/com/defold/extender/services/GCPInstanceService.java
@@ -1,9 +1,14 @@
 package com.defold.extender.services;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/defold/extender/services/GCPInstanceService.java
```diff
@@ -165,12 +170,24 @@ public void checkInstancesState() {
     @PostMapping("/maintenance_mode")
     public void maintenanceMode() {
         LOGGER.info("Set maintenance mode");
+        ExecutorService executor = Executors.newFixedThreadPool(10);
```
britzl, line 173: What is the reason behind the size of 10?

britzl (approved):

server/src/main/java/com/defold/extender/services/GCPInstanceService.java
ekharkunov: Еhe imperial value of how many remote builders need to be "knocked"

PR #708, opened 2025-04-24
Handle glob pattern from Pod resources section

### LINKED ISSUE #296: Handle glob patterns in pod resources
The Cocoapod spec states that a resource can be either a .bundle "file" or individual files. When specifying as individual files the format also supports wildcards. We should use ExtenderUtil.listFilesGlob() when we copy these resources to the bundle output

ekharkunov:
* Handle regex pattern from Pod resources section
* Added test for PodSpec parsing

Fixes #296

### DIFF
--- a/server/src/main/java/com/defold/extender/services/cocoapods/PodSpecParser.java
+++ b/server/src/main/java/com/defold/extender/services/cocoapods/PodSpecParser.java
@@ -5,7 +5,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -663,8 +662,7 @@ public final class PodSpecParser {
      * @param pattern Source file pattern (glob format)
      */
     static void addPodSourceFiles(PodSpec pod, String pattern) {
-        String absolutePatternPath = pod.dir.getAbsolutePath() + File.separator + pattern;
-        List<File> podSrcFiles = listFilesGlob(pod.dir, absolutePatternPath);
+        List<File> podSrcFiles = PodUtils.listFilesGlob(pod.dir, pattern);
         for (File podSrcFile : podSrcFiles) {
             final String filename = podSrcFile.getName();
             if (filename.endsWith(".swift")) {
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/defold/extender/services/cocoapods/PodSpecParser.java
```diff
@@ -760,23 +757,10 @@ static String createOsxUmbrellaHeader(PodSpec pod, File jobDir) throws ExtenderE
         return umbrellaHeaderFile.getAbsolutePath();
     }
 
-    static List<File> listFilesGlob(File dir, String pattern) {
```
ekharkunov, line 763: Moved to separate class.

server/src/main/java/com/defold/extender/services/cocoapods/PodUtils.java
```diff
@@ -0,0 +1,27 @@
...
+import com.defold.extender.ExtenderUtil;
+
+public class PodUtils {
+
+    /*
+     * @dir File Directory where files will be searched
+     * @patter String Path pattern
+    */
+    static List<File> listFilesGlob(File dir, String pattern) {
+        String absPathPattern = Path.of(dir.getAbsolutePath(), pattern).toString();
```
ekharkunov, line 16: Moved common logic which create absolute path pattern here.

server/src/test/java/com/defold/extender/services/cocoapods/CocoaPodsServiceTest.java
```diff
@@ -96,41 +130,29 @@ public void testPodfileParserAll() throws IOException {
         assertArrayEquals(expected, mainPodfile.podnames.toArray());
     }
 
-    @ParameterizedTest
+    @ParameterizedTest(name = "{index}_testParsePodSpecs_{0}")
     @MethodSource("specData")
-    public void testParsePodSpecs(Path spec) throws IOException, ExtenderException {
+    public void testParsePodSpecs(String alias, Path spec) throws IOException, ExtenderException {
         String jsonSpec = Files.readString(spec);
-        File workingDir = Files.createTempDirectory("parse-spec-test").toFile();
```
ekharkunov, line 103: Moved common setup to BeforeEach step

britzl (approved):

PR #811, opened 2025-11-12
Change how contexts are merged

### LINKED ISSUE #771: Ensure consistent application of app manifest
In our implementation, we need to make sure each stage of the manifests are applied correctly.

One use case that doesn't fully work as intended is when a developer wishes to reenable the profiler in a release build:
* Our build_input.yml adds profiler symbols+libs
* Our release_input.appmanifest removes the profiler symbols+libs
* The appmanifest tries to readd the profiler symbols+libs, but they're still not applied

```yaml
      symbols: [ProfilerJS, ProfilerExt]
      excludeLibs: [profile_null, profilerext_null]
      libs: [profile, profilerext, profiler_js]
```

In the final result, the profiler info is not being readded.
It seems the exclude parts of the previous steps are still applied on the appmanifest, which is not how we intended it.

Cons:
We may have situations currently that accidentally rely on this buggy behaviour.

ekharkunov:
Fixes #771 

### Technical details
Change how manifests are applied. Now so-called prune mappings (lists which define what should include or exclude) are taken only from applied manifest. E.g.
```
build.yml <-- ext1.manifest <-- ext2.manifest <-- application.appmanifest.
```

### DIFF
--- a/server/src/main/java/com/defold/extender/ExtenderUtil.java
+++ b/server/src/main/java/com/defold/extender/ExtenderUtil.java
@@ -567,7 +567,10 @@ public class ExtenderUtil
         new PruneMapping("objectFiles", "includeObjectFiles", "excludeObjectFiles"),
         new PruneMapping("dynamicLibs", "includeDynamicLibs", "excludeDynamicLibs"),
         new PruneMapping("symbols", "includeSymbols", "excludeSymbols"),
-        new PruneMapping("jars", "includeJars", "excludeJars"),
+        // if applied manifest contains 'symbols' which were excluded before - need update 'excludeSymbols' list
+        // because depends on that list we defined which extension to build and which symbols should be included into
+        // result binary
+        new PruneMapping("excludeSymbols", "", "symbols"),
         new PruneMapping("frameworks", "includeFrameworks", "excludeFrameworks")
     );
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -560,26 +560,32 @@ private static class PruneMapping {
...
+        new PruneMapping("engineLibs", "includeLibs", "excludeLibs"),
+        new PruneMapping("engineJsLibs", "includeJsLibs", "excludeJsLibs"),
+        new PruneMapping("objectFiles", "includeObjectFiles", "excludeObjectFiles"),
+        new PruneMapping("dynamicLibs", "includeDynamicLibs", "excludeDynamicLibs"),
+        new PruneMapping("symbols", "includeSymbols", "excludeSymbols"),
+        // if applied manifest contains 'symbols' which were excluded before - need update 'excludeSymbols' list
+        // because depends on that list we defined which extension to build and which symbols should be included into
+        // result binary
+        new PruneMapping("excludeSymbols", "", "symbols"),
+        new PruneMapping("frameworks", "includeFrameworks", "excludeFrameworks")
```
ekharkunov: Added missed definition for `frameworks`

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -560,26 +560,32 @@ private static class PruneMapping {
...
-        for (PruneMapping mapping : mappings) {
+        for (PruneMapping mapping : MAPPINGS) {
             List<String> srcList = ExtenderUtil.getStringList(context, mapping.targetName);
             if (srcList.isEmpty())
                 continue;
             context.put(mapping.targetName,
                 ExtenderUtil.pruneItems(srcList,
-                                        ExtenderUtil.getStringList(context, mapping.includeName),
-                                        ExtenderUtil.getStringList(context, mapping.excludeName)) );
+                                        ExtenderUtil.getStringList(b, mapping.includeName),
```
ekharkunov: Main fix: apply prune mappings only from applied context, not from merged and accumulated context.

PR #822, opened 2025-12-03
Update XCode to 26.2, MacOS SDK to 26.2, iOS SDK to 26.2

AGulev:
Build toolchain updated with:  
- XCode 26.2
- macOS SDK 26.2
- iOS SDK 26.2

Also, updated gradle to 9.1.0

Related https://github.com/defold/defold/pull/11563

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -6,7 +6,6 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'io.spring.gradle:dependency-management-plugin:1.1.7'
         classpath 'org.springframework.boot:spring-boot-gradle-plugin:3.5.5'
     }
 }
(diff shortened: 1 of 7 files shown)


server/envs/macos.env
```diff
@@ -5,3 +5,10 @@ XCODE_16_VERSION=16.2
 XCODE_16_CLANG_VERSION=16.0.0
 MACOS_15_VERSION=15.2
 IOS_18_VERSION=18.2
+SWIFT_6_2_VERSION=6.2
+IOS_26_VERSION_MIN=15.0
```
AGulev: min version changed, so we need it in a separate variable now

server/build.gradle
```diff
@@ -29,25 +28,17 @@ ext {
 }
 
 configurations {
-    // A configuration meant for consumers that need the API of this component
-    implementation {
-        canBeResolved = true
-    }
```
AGulev: We don't iterate over implementation artifacts directly, so we don't need it here (it's not supported in new Gradle, so it's easier to just remove it)

PR #923, opened 2026-04-16
Fixed various resource leaks

ekharkunov:


britzl (approved):
Looks good as far as I can tell!

PR #1028, opened 2026-08-19
Replace ProGuard with R8 for Android shrinking

### LINKED ISSUE #566: Replace Proguard with R8
Android developed their own code shrinker R8. So need investigate and make changes to replace usage of Proguard with R8 utility.
https://developer.android.com/build/shrink-code
https://developer.android.com/build/kotlin-support

AGulev:
Android builds can now use R8 for code shrinking and obfuscation, replacing the legacy ProGuard path. R8 preserves Android entry points, extension APIs, dependency consumer rules, and ServiceLoader metadata while producing a mapping file for deobfuscation. Builds without `_app/app.keep` continue to use D8 without shrinking.

Fix https://github.com/defold/extender/issues/566

### Technical changes

- Adds a dedicated R8 pipeline with bounded rule processing, generated keep rules for unconfigured extensions, version-targeted dependency rules, rewritten `META-INF` metadata, dex output, and `mapping.txt`.
- Generates AAPT keep rules for manifest and resource entry points and combines them with application, extension, JAR, and AAR consumer rules.
- Pins R8 8.13.19 in the Android build environment, exposes its path and version through SDK configuration, removes the ProGuard installation, and replaces `DM_DEBUG_DISABLE_PROGUARD` with `DM_DEBUG_DISABLE_R8`.
- Reworks Gradle dependency resolution into one invocation that resolves `releaseRuntimeClasspath`, writes dependency locks and the dependency tree, and emits a deterministic JSON artifact manifest. Gradle is skipped entirely when no dependencies are declared.
- Uses AGP artifact transforms to obtain `EXPLODED_AAR` directories and Jetifier-aware `PROCESSED_JAR` artifacts. Standalone JARs remain unprocessed when Jetifier is disabled, while `android.useAndroidX` stays enabled independently of transformation.
- Removes Extender’s custom `.gradle/unpacked` cache and its extra copy/unpack pass. Artifact paths are canonicalized, validated, deduplicated, and consumed directly from the shared immutable Gradle/AGP cache, reducing duplicate cache storage and filesystem work.
- Supports both traditional exploded AARs and AGP cache layouts, including `jars/classes.jar`, `jars/libs`, resources, assets, JNI libraries, manifests, and consumer rules.
- Preserves legacy Maven resource package names while using collision-safe internal directories for artifacts with identical filenames from different components.
- Keeps D8 as the non-shrinking fallback. Legacy `_app/app.pro` files no longer invoke ProGuard, and the R8 path targets Defold’s API 21+ platform baseline.
- Adds unit and regression coverage for R8 rule selection and safety limits, AAR/JAR layouts, lambdas, ServiceLoader rewriting, Gradle artifact manifests, resource naming, and end-to-end AAR plus classifier-JAR handoff with Jetifier enabled and disabled.

### DIFF
--- a/server/docker/Dockerfile.android.ndk25_sdk36-env
+++ b/server/docker/Dockerfile.android.ndk25_sdk36-env
@@ -13,6 +13,7 @@ ARG ANDROID_NDK_PATH=${ANDROID_ROOT}/android-ndk-r${ANDROID_NDK_VERSION}
 ARG ANDROID_SDK_HOME=${ANDROID_ROOT}/.android
 ARG ANDROID_NDK_BIN_PATH=${ANDROID_NDK_PATH}/toolchains/llvm/prebuilt/linux-x86_64/bin
 ARG ANDROID_SDK_BUILD_TOOLS_PATH=${ANDROID_HOME}/build-tools/${ANDROID_BUILD_TOOLS_VERSION}
+ARG R8_VERSION=8.13.19
 
 FROM europe-west1-docker.pkg.dev/extender-426409/extender-public-registry/extender-build-env:1.1.0 AS build
 
@@ -29,7 +30,7 @@ ARG ANDROID_NDK_VERSION
 ARG ANDROID_NDK_API_VERSION
 ARG ANDROID_64_NDK_API_VERSION
 ARG ANDROID_NDK_FILENAME=android-ndk-r${ANDROID_NDK_VERSION}-linux.tar.gz
-ARG R8_VERSION=8.13.19
+ARG R8_VERSION
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN --mount=type=secret,id=DM_PACKAGES_URL,required=true \
@@ -77,6 +78,7 @@ ARG ANDROID_SDK_BUILD_TOOLS_PATH
 ARG ANDROID_NDK_VERSION
 ARG ANDROID_NDK_API_VERSION
 ARG ANDROID_64_NDK_API_VERSION
+ARG R8_VERSION
 
 ENV ANDROID_ROOT=${ANDROID_ROOT} \
 # ANDROID_HOME has been replaced with ANDROID_SDK_ROOT
@@ -101,21 +103,15 @@ ENV ANDROID_ROOT=${ANDROID_ROOT} \
     ANDROID_NDK_SYSROOT=${ANDROID_NDK_PATH}/toolchains/llvm/prebuilt/linux-x86_64/sysroot \
 # We specify it in build_input.yml by setting it the first in PATH
     PATH=${PATH}:${ANDROID_HOME}/tools:${ANDROID_HOME}/platform-tools:${ANDROID_SDK_BUILD_TOOLS_PATH}:${ANDROID_NDK_BIN_PATH} \
-    ANDROID_PROGUARD=/usr/share/java/proguard.jar
+    ANDROID_R8=${ANDROID_SDK_BUILD_TOOLS_PATH}/lib/d8.jar \
+    ANDROID_R8_VERSION=${R8_VERSION}
 
 COPY --from=build ${ANDROID_ROOT} ${ANDROID_ROOT}
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-# android proguard was version 4.7, this is at least 5.2.1 which seems to work with OpenJDK 11
-RUN \
-  apt-get update && \
-  apt-get install -y --no-install-recommends proguard && \
-  apt-get autoremove -y && \
-  apt-get clean && \
-  rm -rf /var/lib/apt/lists/* && \
 # Since dotnet cannot really cross compile, we need to create a "ar" shim for "llvm-ar"
 # As long as it's in the path, it will be picked up
-  echo '#!/usr/bin/env bash' > /usr/bin/ar && \
+RUN echo '#!/usr/bin/env bash' > /usr/bin/ar && \
   LLVM_AR=$(find /opt/platformsdk/android -iname "llvm-ar" | tail -1) && \
   echo "${LLVM_AR} \$*" >> /usr/bin/ar && \
   chmod +x /usr/bin/ar
--- a/server/src/main/java/com/defold/extender/Extender.java
+++ b/server/src/main/java/com/defold/extender/Extender.java
@@ -69,6 +69,7 @@ class Extender {
     private final TemplateExecutor templateExecutor = new TemplateExecutor();
     private final ProcessExecutor processExecutor = new ProcessExecutor();
     private final ProgressReporter progressReporter;
+    private final R8Configuration r8Configuration;
     private MetricsWriter metricsWriter;
     // context flags
     private Boolean needsCSLibraries = false;
@@ -125,6 +126,7 @@ class Extender {
         Map<String, String> env = new HashMap<String, String>();
         MetricsWriter metricsWriter;
         ProgressReporter progressReporter = ProgressReporter.NOOP;
+        R8Configuration r8Configuration = new R8Configuration();
 
         public Builder() { }
 
@@ -168,6 +170,11 @@ class Extender {
             return this;
         }
 
+        public Builder setR8Configuration(R8Configuration r8Configuration) {
+            this.r8Configuration = r8Configuration;
+            return this;
+        }
+
         public Extender build() throws IOException, ExtenderException {
             return new Extender(this);
         }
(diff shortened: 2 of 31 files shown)


server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -2650,39 +2597,34 @@ private List<File> buildAndroid(String platform) throws ExtenderException {
...
+                (command, commandContext) -> executeCommandLine(command));
+        R8Builder.BuildOutput r8Output = r8Builder.build(
+                allJars,
+                extensionJarMap,
+                mainDexList,
+                aaptKeepRules,
+                aaptMainDexRules);
+        if (r8Output != null) {
+            outputFiles.addAll(Arrays.asList(r8Output.dexFiles));
+            outputFiles.add(r8Output.mappingFile);
```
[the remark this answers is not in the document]
AGulev: fixed

AGulev:
@codex review

server/test-data/sdk/a/defoldsdk/extender/build.yml
```diff
@@ -303,11 +304,12 @@ platforms:
     javacCmd: 'javac -proc:none -source 1.8 -target 1.8 -J-Xms2048m -J-Xmx2048m -classpath {{env.LIBRARYJAR}}:{{classPath}} -d {{classesDir}} @{{sourcesListFile}}'
     jarCmd: 'jar cf {{outputJar}} -C {{classesDir}} .'
     # mainDexList is automatically created by listing all classes inside the engine jars
-    dxCmd: '{{env.ANDROID_BUILD_TOOLS_PATH}}/d8 --main-dex-rules {{mainDexList}} --output {{classes_dex_dir}} --release --lib {{env.LIBRARYJAR}} {{#jars}}{{.}} {{/jars}}'
-    proGuardCmd: 'java -jar {{env.PROGUARD}} {{#src}}-include {{.}} {{/src}} -libraryjars {{env.LIBRARYJAR}} {{#jars}}-injars {{.}} {{/jars}} {{#libraryjars}}-libraryjars {{.}} {{/libraryjars}} -outjar {{tgt}} -printmapping {{mapping}}'
-    proGuardSourceRe: '(?i).+(\.pro)$'
+    dxCmd: '{{env.ANDROID_BUILD_TOOLS_PATH}}/d8 --min-api {{minAndroidSdkVersion}} --main-dex-rules {{mainDexList}} --output {{classes_dex_dir}} --release --lib {{env.LIBRARYJAR}} {{#jars}}{{.}} {{/jars}}'
```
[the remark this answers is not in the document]
AGulev: This appears to conflate R8 with D8. I verified the exact 8.13.19 jar: D8 accepts `--main-dex-rules` with `--min-api` 21, 22, 24, and 35; only R8 rejects it at API 21+, and that path already omits the rules conditionally. Therefore ordinary non-R8 builds do not fail, and this finding is not actionable.

server/src/main/java/com/defold/extender/services/RealGradleService.java
```diff
@@ -116,27 +126,19 @@ public List<File> resolveDependencies(ExtenderBuildState buildState, Map<String,
...
 
     @Override
     public long getCacheSize() throws IOException {
-        Path folder = Paths.get(GRADLE_USER_HOME);
-        return Files.walk(folder)
-          .filter(p -> p.toFile().isFile())
-          .mapToLong(p -> p.toFile().length())
-          .sum();
+        Path folder = Paths.get(this.gradleHome);
+        try (Stream<Path> paths = Files.walk(folder)) {
```
ekharkunov, line 138: 👍

server/src/main/java/com/defold/extender/services/RealGradleService.java
```diff
@@ -208,142 +210,133 @@ private void createBuildGradleFile(File mainGradleFile, List<File> gradleFiles,
...
-            } else if (dependency.endsWith(".jar")) {
-                resolvedDependencies.add(resolveDependencyJAR(file, newName, jobDir));
+
+            String component = (String) componentValue;
+            String originalFileName = (String) originalFileNameValue;
+            String kind = (String) kindValue;
+            File artifact = new File((String) pathValue).getCanonicalFile();
+            GradleArtifact.Kind artifactKind;
+            String resourcePackageName = null;
+            if ("exploded-aar".equals(kind)) {
```
ekharkunov: Is it worth to make strings as constants?

server/docker/Dockerfile.android.ndk25_sdk36-env
```diff
@@ -101,21 +103,15 @@ ENV ANDROID_ROOT=${ANDROID_ROOT} \
     ANDROID_NDK_SYSROOT=${ANDROID_NDK_PATH}/toolchains/llvm/prebuilt/linux-x86_64/sysroot \
 # We specify it in build_input.yml by setting it the first in PATH
     PATH=${PATH}:${ANDROID_HOME}/tools:${ANDROID_HOME}/platform-tools:${ANDROID_SDK_BUILD_TOOLS_PATH}:${ANDROID_NDK_BIN_PATH} \
-    ANDROID_PROGUARD=/usr/share/java/proguard.jar
```
ekharkunov: Seems all old engine's versions start failing after this because Mustache can't render template which contains
```
  android:
    env:
        PROGUARD:                 "{{env.ANDROID_PROGUARD}}"
```

server/src/main/java/com/defold/extender/ExtenderUtil.java
```diff
@@ -929,7 +929,7 @@ public static boolean isMetaInfEntryValuable(ZipEntry entry) {
             && !entryName.endsWith(".MF")
             && !entryName.endsWith(".kotlin_module")
             && !entryName.endsWith(".class")
-            && !entryName.endsWith(".pro")
```
ekharkunov, line 932: Old engines start shipping `.pro` files as part of META-INF here

server/src/main/java/com/defold/extender/R8Builder.java
```diff
@@ -0,0 +1,1024 @@
...
+
+    static final String RULES_WITHOUT_JAR = "r8_rules_without_jar";
+
+    private static final boolean DM_DEBUG_DISABLE_R8 = System.getenv("DM_DEBUG_DISABLE_R8") != null;
+    private static final String APP_RULES_PATH = "_app/app.keep";
+    private static final String JAR_LEGACY_RULE_PREFIX = "META-INF/proguard/";
+    private static final String JAR_R8_RULE_PREFIX = "META-INF/com.android.tools/r8";
+    private static final String GENERATED_EXTENSION_KEEP_ATTRIBUTES =
+            "-keepattributes *Annotation*,Signature,InnerClasses,EnclosingMethod,MethodParameters,Exceptions";
+    static final int MAX_GENERATED_EXTENSION_CLASSES = 32 * 1024;
```
ekharkunov: I think it's worth to move parameters to config file. In case if we need change it in production it's easier and faster change configs instead of rebuilding and redeploying jar

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -301,8 +291,27 @@ private Extender(Builder builder) throws IOException, ExtenderException {
 
             Set<String> keys = this.platformConfig.env.keySet();
             for (String k : keys) {
+                // Older SDKs declare PROGUARD in build.yml. The old command is never
+                // used, so do not require the removed ANDROID_PROGUARD environment.
+                if (k.equals("PROGUARD")) {
```
ekharkunov, line 305: Please, mark it as TODO for further cleanup after 6 month

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -301,8 +291,27 @@ private Extender(Builder builder) throws IOException, ExtenderException {
...
+                    continue;
+                }
+                boolean r8Environment = k.equals("R8") || k.equals("R8_VERSION");
```
ekharkunov: I don't understand why we need that code here. If merged app manifests doesn't contain R8-related stuff than we even doesn't hit this. If R8 stuff presented - we will render environment value in any case. What is the purpose of this code?

server/src/main/java/com/defold/extender/Extender.java
```diff
@@ -419,6 +432,22 @@ private Map<String, Object> createContext(Map<String, Object> src) throws Extend
         return context;
     }
 
+    Map<String, Object> createR8BuilderContext(Map<String, Object> src) throws ExtenderException {
```
ekharkunov: Why we need such processing specifically for R8-related environment?

ekharkunov (changes requested):
Overall good. See comments and suggestions

server/src/main/java/com/defold/extender/Extender.java
AGulev: not needed, removed

server/src/main/java/com/defold/extender/Extender.java
AGulev: not needed

AGulev:
I made fixes  (if I understood all right), thnak you!

ekharkunov (approved):

PR #1040, opened 2026-08-30
Initial SPM support

### LINKED ISSUE #824: Add Swift Package Manager support

ekharkunov:
This PR adds Swift Package Manager support for iOS/macOS extension builds. An extension can now declare SPM dependencies next to (or instead of) CocoaPods, and the resolved products are compiled against and linked into the engine exactly like pod artifacts. CocoaPods and SPM coexist in one build.

## Declaring dependencies

An extension ships a `SwiftPackages.json` in its `ios/` or `osx/` folder:

```json
{
  "platform": "ios",
  "minVersion": "15.0",
  "packages": [
    {
      "url": "https://github.com/getsentry/sentry-cocoa.git",
      "from": "9.0.0",
      "products": ["Sentry"]
    }
  ]
}
```

Each package takes exactly one requirement: `version` (exact), `from`, `branch`, or `revision`. An optional top-level `"wrapperType": "static" | "dynamic"` overrides the wrapper linkage (see below). Manifests from multiple extensions are merged: `minVersion` is the maximum, packages are keyed by canonical URL, and conflicting requirements for the same package fail the build with a clear error.

The manifest is plain JSON — user-authored Swift is never executed. The parser validates and **reconstructs** every value from whitelisted parts (https-only URLs without userinfo/port/query, regex-constrained versions/branches/revisions/product names, size caps), so no raw manifest input ever reaches a generated file or a command line.

## Build flow

For Apple platforms, `AsyncBuilder` invokes the new `SwiftPackageManagerService` (behind `extender.spm.enabled`, off by default) before the extension build:

1. **Discover & parse** — find `SwiftPackages.json` files in the job, parse and merge them (`SpmManifestParser`).
2. **Generate** — render two templates into the job dir:
   * an aggregator `Package.swift` with a single static-library target `SpmDeps` depending on every requested product;
   * an XcodeGen `project.yml` for a wrapper project with one framework target `SpmWrapper` that links the `SpmDeps` package product.
3. **Build** — run `xcodegen generate`, then `xcodebuild -project SpmWrapper.xcodeproj -scheme SpmWrapper -configuration Release build` with a destination mapped from the target platform/arch. Xcode resolves the package graph, fetches sources/binary artifacts, and compiles everything. Full output is captured to `build.log` in the job folder.
4. **Harvest** — locate the wrapper's final link line in the log (`SpmBuildOutputParser`; clang `-dynamiclib` line for the dynamic shape, `Libtool -static` for the static one) and combine it with the products directory (`ResolvedPackages.harvest`):
   * built frameworks, system frameworks/libs, framework/library search paths, rpaths;
   * resource bundles and privacy manifests (deduped by canonical path);
   * include paths for the extension's own sources, derived from the generated module maps (so `#import <Module/Header.h>` and `<Module>-Swift.h` work for source-built packages);
   * the Swift runtime lib dir for the static shape;
   * `Package.resolved` as the lock file emitted with the build output.
5. **Consume** — the harvested result feeds the normal compile/link flow (see integration below); dynamic frameworks are copied to the build output for embedding.

## Static vs dynamic wrapper

By default the wrapper builds as `MACH_O_TYPE=staticlib`: libtool folds all package objects into one archive that links straight into the engine — no embedding, no load-time cost. Some SDKs don't survive static merging (e.g. binary xcframeworks whose duplicate inputs libtool folds, producing duplicate symbols under the engine's `-ObjC`); a manifest can opt into `"wrapperType": "dynamic"`, which builds an `mh_dylib` wrapper that is embedded into the app bundle with the matching rpath flags. Both shapes are first-class; `extender.spm.wrapper-mach-o-type` sets the default.

## Integration: `ResolvedNativeDeps`

The CocoaPods-specific plumbing in `Extender` is generalized behind a new `ResolvedNativeDeps` interface (frameworks, search paths, static libs, include paths, link flags, min OS version, resources/bundles, dynamic frameworks, lock file, privacy manifests). `ResolvedPods` implements it via delegating aliases — no CocoaPods behavior or test changes — and the new `ResolvedPackages` is the SPM implementation. `Extender` iterates a `List<ResolvedNativeDeps>`, so compile/link argument collection, resource copying, framework embedding, and lock-file emission are shared between both dependency managers. The platform min version used for the engine link is the max across all resolvers.

## Multi-Xcode & caching

One Extender instance can serve jobs pinned to different Xcode versions, so `DEVELOPER_DIR` is resolved **per build**: the job's `XCODE_VERSION` is looked up in the `extender.spm.xcode-developer-dirs` map, falling back to `extender.spm.default-developer-dir`. Selection is per-process env only — never `xcode-select` — so concurrent jobs on different Xcodes don't interfere.

DerivedData, the module cache, and `-clonedSourcePackagesDirPath` are per-job (SwiftPM prunes the cloned-packages dir to the current graph, so it must not be shared). Cross-job warmth comes from a shared `-packageCachePath` (git mirrors + binary artifacts), keyed by Xcode version and rotated/cleaned by crons mirroring the CocoaPods cache machinery.

## Hardening

`xcodebuild` runs with `-scmProvider system` and `GIT_CONFIG_NOSYSTEM=1`, `GIT_CONFIG_GLOBAL=/dev/null`, `GIT_TERMINAL_PROMPT=0`, `GIT_ASKPASS=/usr/bin/true`: package fetching uses system git with no user/system config, no credential prompts, and no keychain-backed git credential lookups. Code signing is disabled (`CODE_SIGNING_ALLOWED=NO`) and package plugins are not validated interactively (`-skipPackagePluginValidation`). Note: packages shipping **binary artifacts** hit SwiftPM's internal downloader, whose keychain-backed auth provider has no off-switch — on a fresh builder the keychain must either contain no items for the artifact hosts or have access pre-granted once.

## Configuration

```yaml
extender:
    spm:
        enabled: false
        home-dir-prefix: /tmp/.swiftpm
        default-developer-dir: /Applications/Xcode.app/Contents/Developer
        # xcode-developer-dirs:
        #     "26.2": /Applications/Xcode_26.2.app/Contents/Developer
        xcodegen-path: xcodegen
        wrapper-mach-o-type: staticlib   # staticlib | mh_dylib
        swift-version: "6.0"
        cache-dir-rotate-cron: "0 20 2 * * *"
        old-cache-clean-cron: "0 20 6 * * *"
```

Enabled in the `standalone-dev` profile. Builders additionally need `xcodegen` installed. A new `extender.job.spm.resolve` metric times the resolution stage.

Fixes #824


ekharkunov:
@codex review

britzl (approved):
Looks good as far as I can tell!

AGulev:
we should update extensions manuals as well

